PR #1, opened 2019-06-19
SFX-94: Set up directories and build tools within Logic layer repository

JoeyDeol:


packages/@sfx/sayt-plugin/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "@sfx/sayt-plugin",
+  "version": "0.0.0",
+  "description": "",
+  "main": "src/index.ts",
+  "scripts": {
+    "build": "../../../scripts/build.sh",
+    "dev": "nodemon --watch src --exec npm run build",
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
```
juliamlim: Let's throw this in for reference as per https://docs.npmjs.com/files/package.json#repository
```suggestion
  },
  "repository": {
  "type" : "git",
  "url" : "https://github.com/groupby/sfx-logic.git",
  "directory": "packages/@sfx/sayt-plugin"
},
```

webpack.config.js
```diff
@@ -0,0 +1,51 @@
...
+    filename: '[name].js',
+    path: path.resolve(__dirname, 'dist')
+  },
+
+  plugins: [new webpack.ProgressPlugin()],
+
+  module: {
+    rules: [{
+      test: /.(ts|tsx)?$/,
+      loader: 'ts-loader',
```
juliamlim: This is just a suggestion but do we want to look into `awesome-typescript-loader`, this way we can just point all components up to the main tsconfig flies rather than duplicating the same file in every package.

packages/@sfx/sayt-plugin/package.json
JoeyDeol: Updated.

PR #4, opened 2019-06-26
SFX-95: Testing Setup

geoffcoutts:
DESCRIPTION: 
- Setup for testing suite
  - Update dependencies
  - Update package scripts for running tests in browser and node
- Add sample module and test
- Update README with information for tests


karma.conf.js
```diff
@@ -0,0 +1,84 @@
+// Karma configuration
```
jmbtutor: We've changed this file significantly enough that I think we should clean it up.

tsconfig.json
```diff
@@ -7,6 +7,7 @@
       "dom"
     ],
     "types": [
+      "mocha",
```
jmbtutor: We should keep the order consistent between all the tsconfigs. It'll make scripting changes much easier.

README.md
```diff
@@ -30,6 +30,28 @@ To build an individual package in response to changes within the `src` directory
...
+
+- Chrome
+
+To run tests for a specific plugin package, navigate to its directory and use the following commands based on the desired testing flow:
+
+- To run the tests for a specific package once:
+```sh
+yarn test
+```
+- To watch the `./src` and `./test` directories and run the tests after any changes:
```
jmbtutor: I would remove the `./` from these because they're not relative to this file.

README.md
```diff
@@ -30,6 +30,28 @@ To build an individual package in response to changes within the `src` directory
...
+```sh
+yarn tdd
+```
+- To run the tests in a browser environment:
+```sh
+yarn test:browser
```
jmbtutor: Do we have a watch equivalent for this? Or if this is a watch, do we have a run-once equivalent?

karma.conf.js
geoffcoutts: Would you propose removing the comments, superfluous lines, and alphabetizing the options?

.mocharc.json
```diff
@@ -0,0 +1,8 @@
+{
```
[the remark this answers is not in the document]
geoffcoutts: The view layer is not using a `.mocharc.json`.

karma.conf.js
jmbtutor: I would say to remove comments that don't really add any clarity (which looks to be all of the ones that are in here), superfluous lines (like collapsing multiple empty lines into one), but I don't think we need to alphabetize the options as long as the order makes sense.

.mocharc.json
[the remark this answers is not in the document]
jmbtutor: Great, then there's nothing left to do with this one. I'll resolve this.

packages/@sfx/sayt-plugin/src/index.ts
```diff
@@ -3,14 +3,18 @@
  */
 const testObj = {
   label: 'test text',
+  number: 6,
 }
 
-const testFunc = (text: TestInterface) => {
-  console.log(`I am logging out ${ text.label }`);
+export const testFunc = (data: DummyInterface) => {
+  return exports.getLabel(data).toUpperCase();
```
jmbtutor: I'm not sure we should be doing this, but we're not going to be keeping this, so I'll ignore it for now.

packages/@sfx/sayt-plugin/test/unit/common/index.test.ts
```diff
@@ -0,0 +1,18 @@
...
+    label: 'test',
+    number: 6,
+  };
+  it('should return the label string capitalized', () => {
+    expect(Sample.testFunc(testData)).to.equal('TEST');
+  });
+  it('should call a function to get the label value', () => {
+    const getLabelStub = stub(Sample, 'getLabel').callThrough();
+    Sample.testFunc(testData);
+    expect(getLabelStub.callCount).to.equal(1);
```
jmbtutor: Does `expect(getLabelStub).to.be.calledOnce` works? This way, we can test that sinon-chai works.

geoffcoutts: Yup, will change to that. Completely blanked about writing it out that way.

packages/@sfx/sayt-plugin/test/unit/common/index.test.ts
```diff
@@ -0,0 +1,20 @@
+import { expect, stub } from '../../utils';
```
DanielMoniz-GBI: This was a source of great pain for me. I didn't manage to get the same file running in Karma and Mocha.

test/setup.ts
```diff
@@ -0,0 +1,8 @@
+import { chai, sinon } from './utils';
+import * as sinonChai from 'sinon-chai';
+
+chai.use(sinonChai);
+
+afterEach(() => {
+  sinon.restore();
```
DanielMoniz-GBI: I believe sinon.restore() is deprecated. People are now encouraged to use sandboxes.

https://github.com/sinonjs/sinon/issues/1390

test/utils.ts
```diff
@@ -0,0 +1,10 @@
+import * as sinonImport from 'sinon';
+import * as chaiImport from 'chai';
+
+declare var global: any;
+
+export const sinon = global.sinon || sinonImport;
+export const chai = global.chai || chaiImport;
```
DanielMoniz-GBI: Can you explain the idea with the above four lines?

packages/@sfx/sayt-plugin/test/unit/common/index.test.ts
geoffcoutts: Plenty of issues.... Every time we added a new tool for testing it would break.

test/setup.ts
geoffcoutts: Starting with sinon v5.0, sinon itself becomes the default sandbox. 

https://sinonjs.org/guides/migrating-to-5.0.html

test/utils.ts
geoffcoutts: This setup file will make sinon, chai, and some of their methods available for the tests in the global scope as it is run by karma and mocha before the test files. We were having difficulties getting karma specifically to pick them up.

PR #5, opened 2019-07-03
SFX-99: Add Core module with registration mechanism

jmbtutor:
Core is the backbone of the entire SF-X system. It manages plugins and allows them to communicate with each other. The core package defines interfaces for plugins.

A registration mechanism is currently implemented. Unregistration will be added separately.

### DIFF
--- a/packages/@sfx/core/README.md
+++ b/packages/@sfx/core/README.md
@@ -1,1 +1,27 @@
 # SF-X Core
+
+This package contains the SF-X Core class and related plugin interfaces.
+
+## Usage
+
+To use Core, simply instantiate it:
+
+```js
+const core = new Core();
+```
+
+## Registering a plugin
+
+To register one or more plugins with Core, instantiate the plugins, then
+pass them to `Core.register()`:
+
+```js
+const core = new Core();
+const pluginA = new PluginA();
+const pluginB = new PluginB();
+
+core.register([pluginA, pluginB]);
+```
+
+A plugin may take configuration options in its constructor. Refer to the
+plugin's documentation for details.
--- a/packages/@sfx/core/README.md
+++ b/packages/@sfx/core/README.md
@@ -1,0 +1,1 @@
+# SF-X Core
--- a/packages/@sfx/core/src/core.ts
+++ b/packages/@sfx/core/src/core.ts
@@ -1,2 +1,14 @@
+import { Plugin } from './plugin';
+
 export default class Core {
+  plugins = Object.create(null);
+
+  /**
+   * Register one or more plugins with Core.
+   *
+   * @param plugins An array of plugin instances to regsiter.
+   */
+  register(plugins: Plugin[]) {
+    plugins.forEach((p) => p.register(this.plugins));
+  }
 }
--- a/packages/@sfx/core/src/core.ts
+++ b/packages/@sfx/core/src/core.ts
@@ -6,11 +6,43 @@ import {
   registerPlugins,
 } from './utils/core';
 
+/**
+ * The core of the SF-X plugin system. This entity is responsible for
+ * managing plugins and provides a mechanism for plugins to communicate
+ * with each other.
+ */
 export default class Core {
+  /**
+   * The plugin registry. This object is a dictionary containing plugin
+   * names as keys and their corresponding exposed values as values.
+   *
+   * The preferred way to access plugins in the registry is through
+   * another plugin. Accessing the plugins through Core outside of a
+   * plugin is discouraged.
+   */
   plugins: PluginRegistry = Object.create(null);
 
   /**
    * Register one or more plugins with Core.
+   *
+   * Plugins given here are registered as a batch: the next lifecycle
+   * event does not occur until all plugins in this batch have
+   * experienced the current lifecycle event.
+   *
+   * The registration lifecycle is as follows:
+   *
+   * 1. **Registration:** The registry is populated with each plugin's
+   *    name and the value that it wants to expose. Plugins cannot
+   *    assume that other plugins have been registered.
+   * 2. **Initialization:** Plugins perform setup tasks. Plugins are
+   *    aware of the existence of other plugins, but should not make use
+   *    of their functionality as they may not yet be initialized.
+   * 3. **Ready:** Plugins may make use of other plugins.
+   *
+   * This function ensures that all plugin dependencies are available
+   * before proceeding to register the plugins. Circular dependencies
+   * are supported. If dependencies are missing, an error will be
+   * thrown.
    *
    * @param plugins An array of plugin instances to regsiter.
    */
--- a/packages/@sfx/core/src/core.ts
+++ b/packages/@sfx/core/src/core.ts
@@ -1,4 +1,5 @@
 import { Plugin } from './plugin';
+import { calculateMissingDependencies } from './utils/core';
 
 export default class Core {
   plugins = Object.create(null);
@@ -9,6 +10,10 @@ export default class Core {
    * @param plugins An array of plugin instances to regsiter.
    */
   register(plugins: Plugin[]) {
+    const missingDependencies = calculateMissingDependencies(plugins, this.plugins);
+    if (missingDependencies.length) {
+      throw new Error('Missing dependencies: ' + missingDependencies.join(', '));
+    }
     // ensure dependencies are valid
       // if not, exit/throw
     // call lifecycle methods on plugins
--- a/packages/@sfx/core/src/core.ts
+++ b/packages/@sfx/core/src/core.ts
@@ -17,10 +17,10 @@ export default class Core {
    * names as keys and their corresponding exposed values as values.
    *
    * The preferred way to access plugins in the registry is through
-   * another plugin. Accessing the plugins through Core outside of a
+   * another plugin. Accessing the registry through Core outside of a
    * plugin is discouraged.
    */
-  plugins: PluginRegistry = Object.create(null);
+  registry: PluginRegistry = Object.create(null);
 
   /**
    * Register one or more plugins with Core.
@@ -47,12 +47,12 @@ export default class Core {
    * @param plugins An array of plugin instances to register.
    */
   register(plugins: Plugin[]) {
-    const missingDependencies = calculateMissingDependencies(plugins, this.plugins);
+    const missingDependencies = calculateMissingDependencies(plugins, this.registry);
     if (missingDependencies.length) {
       throw new Error('Missing dependencies: ' + missingDependencies.join(', '));
     }
 
-    registerPlugins(plugins, this.plugins);
+    registerPlugins(plugins, this.registry);
     initPlugins(plugins);
     readyPlugins(plugins);
   }
--- a/packages/@sfx/core/src/core.ts
+++ b/packages/@sfx/core/src/core.ts
@@ -37,7 +37,8 @@ export default class Core {
    * 2. **Initialization:** Plugins perform setup tasks. Plugins are
    *    aware of the existence of other plugins, but should not make use
    *    of their functionality as they may not yet be initialized.
-   * 3. **Ready:** Plugins may make use of other plugins.
+   * 3. **Ready:** Registration is complete. All plugins have been
+   *    initialized, and plugins may make use of other plugins.
    *
    * This function ensures that all plugin dependencies are available
    * before proceeding to register the plugins. Circular dependencies
--- a/packages/@sfx/core/src/core.ts
+++ b/packages/@sfx/core/src/core.ts
@@ -1,0 +1,2 @@
+export default class Core {
+}
--- a/packages/@sfx/core/src/core.ts
+++ b/packages/@sfx/core/src/core.ts
@@ -1,4 +1,4 @@
-import { Plugin } from './plugin';
+import { Plugin, PluginRegistry } from './plugin';
 import {
   calculateMissingDependencies,
   initPlugins,
@@ -7,7 +7,7 @@ import {
 } from './utils/core';
 
 export default class Core {
-  plugins = Object.create(null);
+  plugins: PluginRegistry = Object.create(null);
 
   /**
    * Register one or more plugins with Core.
--- a/packages/@sfx/core/src/core.ts
+++ b/packages/@sfx/core/src/core.ts
@@ -1,5 +1,10 @@
 import { Plugin } from './plugin';
-import { calculateMissingDependencies } from './utils/core';
+import {
+  calculateMissingDependencies,
+  initPlugins,
+  readyPlugins,
+  registerPlugins,
+} from './utils/core';
 
 export default class Core {
   plugins = Object.create(null);
@@ -14,8 +19,9 @@ export default class Core {
     if (missingDependencies.length) {
       throw new Error('Missing dependencies: ' + missingDependencies.join(', '));
     }
-    // ensure dependencies are valid
-      // if not, exit/throw
-    // call lifecycle methods on plugins
+
+    registerPlugins(plugins, this.plugins);
+    initPlugins(plugins);
+    readyPlugins(plugins);
   }
 }
--- a/packages/@sfx/core/src/core.ts
+++ b/packages/@sfx/core/src/core.ts
@@ -9,6 +9,8 @@ export default class Core {
    * @param plugins An array of plugin instances to regsiter.
    */
   register(plugins: Plugin[]) {
-    plugins.forEach((p) => p.register(this.plugins));
+    // ensure dependencies are valid
+      // if not, exit/throw
+    // call lifecycle methods on plugins
   }
 }
--- a/packages/@sfx/core/src/core.ts
+++ b/packages/@sfx/core/src/core.ts
@@ -44,7 +44,7 @@ export default class Core {
    * are supported. If dependencies are missing, an error will be
    * thrown.
    *
-   * @param plugins An array of plugin instances to regsiter.
+   * @param plugins An array of plugin instances to register.
    */
   register(plugins: Plugin[]) {
     const missingDependencies = calculateMissingDependencies(plugins, this.plugins);
--- a/packages/@sfx/core/src/plugin.ts
+++ b/packages/@sfx/core/src/plugin.ts
@@ -1,6 +1,7 @@
 export interface Plugin {
   metadata: PluginMetadata;
   register(plugins: object): any;
+  init?: () => void;
 }
 
 export interface PluginMetadata {
--- a/packages/@sfx/core/src/plugin.ts
+++ b/packages/@sfx/core/src/plugin.ts
@@ -1,3 +1,8 @@
 export interface Plugin {
+  metadata: PluginMetadata;
   register(plugins: object): any;
+}
+
+export interface PluginMetadata {
+  name: string;
 }
--- a/packages/@sfx/core/src/plugin.ts
+++ b/packages/@sfx/core/src/plugin.ts
@@ -1,0 +1,3 @@
+export interface Plugin {
+  register(plugins: object): any;
+}
--- a/packages/@sfx/core/src/plugin.ts
+++ b/packages/@sfx/core/src/plugin.ts
@@ -48,7 +48,7 @@ export interface Plugin {
    * The callback for the ready phase of the lifecycle.
    *
    * In this function, the plugin can assume that other plugins have
-   * been initialized.
+   * been initialized and may safely use other plugins.
    */
   ready?: () => void;
 }
--- a/packages/@sfx/core/src/plugin.ts
+++ b/packages/@sfx/core/src/plugin.ts
@@ -1,6 +1,6 @@
 export interface Plugin {
   metadata: PluginMetadata;
-  register: (plugins: object) => any;
+  register: (plugins: PluginRegistry) => any;
   init?: () => void;
   ready?: () => void;
 }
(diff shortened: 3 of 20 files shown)


packages/@sfx/core/src/core.ts
```diff
@@ -0,0 +1,59 @@
...
+export default class Core {
+  /**
+   * The plugin registry. This object is a dictionary containing plugin
+   * names as keys and their corresponding exposed values as values.
+   *
+   * The preferred way to access plugins in the registry is through
+   * another plugin. Accessing the plugins through Core outside of a
+   * plugin is discouraged.
+   */
+  plugins: PluginRegistry = Object.create(null);
```
jmbtutor: Should this be called `registry` instead?

packages/@sfx/core/test/unit/common/utils/core.test.ts
```diff
@@ -0,0 +1,315 @@
+import { expect, spy } from '../../../utils';
```
geoffcoutts: It looks like this file is not being run in the node testing environment as the `.mocharc.json` is not set up to watch subdirectories inside of the common/node directories.

packages/@sfx/core/test/unit/common/core.test.ts
```diff
@@ -0,0 +1,84 @@
+import { expect, sinon, spy, stub } from '../../utils';
```
geoffcoutts: `spy` is not being used in the file and can be removed from the import. 
```suggestion
import { expect, sinon, stub } from '../../utils';
```

packages/@sfx/core/test/unit/common/core.test.ts
```diff
@@ -0,0 +1,84 @@
...
+      expect(core.plugins).to.be.empty;
+      expect(Object.getPrototypeOf(core.plugins)).to.be.null;
+    });
+  });
+
+  describe('register()', () => {
+    let calculateMissingDependencies = stub(CoreUtils, 'calculateMissingDependencies').returns(['x']);
+    let registerPlugins = stub(CoreUtils, 'registerPlugins');
+    let initPlugins = stub(CoreUtils, 'initPlugins');
+    let readyPlugins = stub(CoreUtils, 'readyPlugins');
```
geoffcoutts: The above 4 stubs seem redundant as this is also being done in the `beforeEach` step. Could you just initialize these first without the stub?

packages/@sfx/core/test/unit/common/index.test.ts
```diff
@@ -0,0 +1,34 @@
+import { expect } from '../../utils';
+import {
+  Core as CoreExport,
+  Plugin as PluginExport,
+  PluginMetadata as PluginMetadataExport,
+  PluginRegistry as PluginRegistryExport,
+} from '../../../src';
+import Core from '../../../src/core';
+import { Plugin, PluginMetadata, PluginRegistry } from '../../../src/plugin';
+import * as CoreUtils from '../../../src/utils/core';
```
geoffcoutts: `CoreUtils` is not being used in this file.

packages/@sfx/core/src/utils/core.ts
```diff
@@ -0,0 +1,79 @@
...
+ * @returns An array of names of missing plugins.
+ */
+export function calculateMissingDependencies(plugins: Plugin[], registry: object): string[] {
+  const available = [
+    ...Object.keys(registry),
+    ...plugins.map(({ metadata: { name }}) => name),
+  ];
+  const availableSet = new Set(available);
+
+  const required = plugins.reduce((memo, plugin) => {
```
geoffcoutts: `reduce` seems like the wrong method to use on the array here as the return value isn't always being reduced in size compared to the inputs (ignoring overall structure of the plugins array). While both methods can technically work, wouldn't `map` be the more correct method?

packages/@sfx/core/src/plugin.ts
```diff
@@ -0,0 +1,83 @@
...
+   */
+  depends: string[];
+}
+
+/**
+ * The type of the plugin registry. Each key/value pair corresponds to
+ * the name of a plugin and its exposed value.
+ */
+export interface PluginRegistry {
+  [key: string]: any;
```
DanielMoniz-GBI: JavaScript object keys are all strings. Do we need to specify a type here?

packages/@sfx/core/src/utils/core.ts
DanielMoniz-GBI: `reduce()` is being used because each plugin in the array actually contains an array of dependencies. This method is essentially just flattening those arrays. There may be a way to use `Array.prototype.flat()` in combination with `map()`, here, but I don't know that it will be any simpler.

packages/@sfx/core/src/utils/core.ts
```diff
@@ -0,0 +1,79 @@
...
+  });
+
+  Object.assign(registry, newlyRegistered);
+
+  return newlyRegistered;
+}
+
+/**
+ * Calls the `init` function of each plugin. If a plugin does not
+ * exppose an `init` function, it is ignored.
```
JoeyDeol:
```suggestion
 * expose an `init` function, it is ignored.
```

packages/@sfx/core/src/utils/core.ts
```diff
@@ -0,0 +1,79 @@
...
+  ];
+  const availableSet = new Set(available);
+
+  const required = plugins.reduce((memo, plugin) => {
+    return [...memo, ...plugin.metadata.depends];
+  }, []);
+  const requiredSet = new Set(required);
+
+  const difference = new Set(Array.from(requiredSet).filter((p) => !availableSet.has(p)));
+
```
DanielMoniz-GBI: We can remove some of the empty lines above.

packages/@sfx/core/src/utils/core.ts
JoeyDeol: Also you should consider rewording "it is ignored". It is unclear what you mean.

packages/@sfx/core/src/utils/core.ts
```diff
@@ -0,0 +1,79 @@
...
+  plugins.forEach((plugin) => {
+    const exposedValue = plugin.register(Object.create(registry));
+    const { name } = plugin.metadata;
+
+    newlyRegistered[name] = exposedValue;
+  });
+
+  Object.assign(registry, newlyRegistered);
+
```
DanielMoniz-GBI: You can probably just remove this newline.

packages/@sfx/core/src/utils/core.ts
DanielMoniz-GBI:
```suggestion
 * expose an `init` function, it is ignored.
```

packages/@sfx/core/src/utils/core.ts
```diff
@@ -0,0 +1,79 @@
...
+  plugins.forEach((plugin) => {
+    if (typeof plugin.init === 'function') {
+      plugin.init();
+    }
+  });
+}
+
+/**
+ * Calls the `ready` function of each plugin. If a plugin does not
+ * expose a `ready` function, it is ignored.
```
JoeyDeol: Same as above, consider rewording "it is ignored."

packages/@sfx/core/test/unit/common/core.test.ts
```diff
@@ -0,0 +1,84 @@
+import { expect, sinon, spy, stub } from '../../utils';
+import Core from '../../../src/core';
+import * as CoreUtils from '../../../src/utils/core';
```
DanielMoniz-GBI: I'd love to get `src/` on the path so that modules can simply start there. The above can cause issues if/when refactoring the directory structure needs to happen.

yarn.lock
```diff
@@ -4674,6 +4674,11 @@ sinon-chai@^3.3.0:
   resolved "https://registry.yarnpkg.com/sinon-chai/-/sinon-chai-3.3.0.tgz#8084ff99451064910fbe2c2cb8ab540c00b740ea"
   integrity sha512-r2JhDY7gbbmh5z3Q62pNbrjxZdOAjpsqW/8yxAZRSqLZqowmfGZPGUZPFf3UX36NLis0cv8VEM5IJh9HgkSOAA==
 
+sinon-chai@^3.3.0:
+  version "3.3.0"
+  resolved "https://registry.yarnpkg.com/sinon-chai/-/sinon-chai-3.3.0.tgz#8084ff99451064910fbe2c2cb8ab540c00b740ea"
+  integrity sha512-r2JhDY7gbbmh5z3Q62pNbrjxZdOAjpsqW/8yxAZRSqLZqowmfGZPGUZPFf3UX36NLis0cv8VEM5IJh9HgkSOAA==
+
```
JoeyDeol: Should this have been introduced after Geoff's PR?

packages/@sfx/core/test/unit/common/core.test.ts
```diff
@@ -0,0 +1,84 @@
...
+        },
+      ];
+      calculateMissingDependencies.returns(['x']);
+      Object.assign(core.plugins, { m: 'mm' });
+
+      expect(() => core.register(plugins)).to.throw();
+      expect(calculateMissingDependencies).to.be.calledWith(plugins, sinon.match(core.plugins));
+    });
+
+    it('should call the lifecycle events on the plugins in order', () => {
```
DanielMoniz-GBI: To be pedantic about the language clarity:
```suggestion
    it('should call the lifecycle events in order on the plugins', () => {
```

packages/@sfx/core/test/unit/common/core.test.ts
```diff
@@ -0,0 +1,84 @@
...
+            depends: [],
+          },
+        },
+      ];
+      calculateMissingDependencies.returns([]);
+
+      core.register(plugins);
+
+      expect(registerPlugins).to.be.calledWith(plugins, sinon.match(core.plugins));
+      expect(registerPlugins).to.be.calledWith(plugins, sinon.match(core.plugins));
```
JoeyDeol: Repeat line, please remove.

packages/@sfx/core/test/unit/common/index.test.ts
DanielMoniz-GBI: He's not the import referee we deserved. He's the import referee we needed.

packages/@sfx/core/src/utils/core.ts
jmbtutor: `reduce` is correct here, though I admit that there are more performant (though possibly less readable) ways of achieving this. The trick is to realize that `plugin.metadata.depends` is an array of strings and not simply a string, so this `reduce` invocation actually flattens the arrays into a single array. `map` wouldn't be able to do that without being misused.

packages/@sfx/core/test/unit/common/index.test.ts
```diff
@@ -0,0 +1,34 @@
...
+  it('should export the PluginMetadata interface', () => {
+    // The following should cause a compiler error if the two interfaces are not equal.
+    const plugin_metadata_is_assignable_to_plugin_metadata_export: PluginMetadataExport = {} as PluginMetadata;
+    const plugin_metadata_export_is_assignable_to_plugin_metadata: PluginMetadata = {} as PluginMetadataExport;
+  });
+
+  it('should export the PluginRegistry interface', () => {
+    // The following should cause a compiler error if the two interfaces are not equal.
+    const plugin_registry_is_assignable_to_plugin_registry_export: PluginRegistryExport = {} as PluginRegistry;
+    const plugin_registry_export_is_assignable_to_plugin_registry: PluginRegistry = {} as PluginRegistryExport;
```
DanielMoniz-GBI: Any reason we're not using camelCase in this file?

packages/@sfx/core/test/unit/common/utils/core.test.ts
```diff
@@ -0,0 +1,315 @@
...
+        {
+          metadata: {
+            name: 'n',
+            depends: ['z'],
+          },
+        },
+      ]
+
+      const missing = calculateMissingDependencies(plugins, registry);
+
```
DanielMoniz-GBI: You can remove this newline, and consider doing away with the one-use variable.

packages/@sfx/core/test/unit/common/utils/core.test.ts
```diff
@@ -0,0 +1,315 @@
...
+        {
+          metadata: {
+            name: 'n',
+            depends: ['b'],
+          },
+        },
+      ]
+
+      const missing = calculateMissingDependencies(plugins, registry);
+
```
DanielMoniz-GBI: Same as above.

packages/@sfx/core/test/unit/common/utils/core.test.ts
```diff
@@ -0,0 +1,315 @@
...
+        {
+          metadata: {
+            name: 'n',
+            depends: [],
+          },
+        },
+      ];
+
+      const missing = calculateMissingDependencies(plugins, registry);
+
```
DanielMoniz-GBI: Same as above.

packages/@sfx/core/test/unit/common/utils/core.test.ts
```diff
@@ -0,0 +1,315 @@
...
+        {
+          metadata: {
+            name: 'b',
+            depends: ['c'],
+          },
+        },
+      ];
+
+      const missing = calculateMissingDependencies(plugins, registry);
+
```
DanielMoniz-GBI: Same as above.

packages/@sfx/core/test/unit/common/core.test.ts
jmbtutor: Looks like this was a leftover from an incomplete refactor. You're right, these should be removed and the ones in the `beforeEach` should stay.

packages/@sfx/core/test/unit/common/utils/core.test.ts
```diff
@@ -0,0 +1,315 @@
...
+        {
+          metadata: {
+            name: 'c',
+            depends: ['a'],
+          },
+        },
+      ];
+
+      const missing = calculateMissingDependencies(plugins, registry);
+
```
DanielMoniz-GBI: Same as above.

packages/@sfx/core/test/unit/common/utils/core.test.ts
```diff
@@ -0,0 +1,315 @@
...
+      const cvalue = registry.c = 'cc';
+
+      expect(localRegistryA.c).to.equal(cvalue);
+      expect(localRegistryB.c).to.equal(cvalue);
+    });
+
+    it('should return a map of new plugin keys to exposed values', () => {
+      const valueA = { a: 'a' };
+      const valueB = { b: 'b' };
+      const valueC = { c: 'c' };
```
DanielMoniz-GBI: We probably should have considered using varying datatypes for the values above. It makes it look like plugins have to expose objects, which is not true, and I find that distracts from the test.

packages/@sfx/core/src/utils/core.ts
geoffcoutts: Yup, I missed that part when I read through the code. Thanks for the clarification.

packages/@sfx/core/src/core.ts
geoffcoutts: If `plugins` has been renamed to `registry`, should the interface be renamed as well?

packages/@sfx/core/src/core.ts
```diff
@@ -0,0 +1,59 @@
...
+   *
+   * The registration lifecycle is as follows:
+   *
+   * 1. **Registration:** The registry is populated with each plugin's
+   *    name and the value that it wants to expose. Plugins cannot
+   *    assume that other plugins have been registered.
+   * 2. **Initialization:** Plugins perform setup tasks. Plugins are
+   *    aware of the existence of other plugins, but should not make use
+   *    of their functionality as they may not yet be initialized.
+   * 3. **Ready:** Plugins may make use of other plugins.
```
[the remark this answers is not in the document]
JoeyDeol: Rather I meant to say that we should mention that all plugins should have been initialized and their functionality is now available for use within the ready phase.

yarn.lock
JoeyDeol: Cool sounds good, resolving.

packages/@sfx/core/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+To register one or more plugins with Core, instantiate the plugins, then
+pass them to `Core.register()`:
+
+```js
+const core = new Core();
+const pluginA = new PluginA();
+const pluginB = new PluginB();
+
+core.register([pluginA, pluginB]);
```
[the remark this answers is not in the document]
jmbtutor: Accepting an array allows us to accept registration options that apply to the batch in the future. Looking ahead to a possible "provides" implementation, one such option may be to specify which plugin, out of multiple plugins that provide the same name, should be exposed under that name.

packages/@sfx/core/src/plugin.ts
jmbtutor: Unfortunately, yes.

packages/@sfx/core/src/core.ts
jmbtutor: The interface name is already `PluginRegistry`. Since it will be used outside of `Core`, I think it can stay the way it is.

packages/@sfx/core/test/unit/common/index.test.ts
jmbtutor: These variables aren't really being used for anything, but rather as descriptions for what's being tested. I thought it was more readable to use snake_case here, but if you want me to move the descriptions to a comment instead and call the variables something more generic, I can do that.

packages/@sfx/core/test/unit/common/utils/core.test.ts
jmbtutor: The idea with having this newline and the one-use variable is to split the test into three parts: [Arrange, Act, and Assert](http://wiki.c2.com/?ArrangeActAssert). I would rather optimize for clarity than performance in the tests.

packages/@sfx/core/test/unit/common/utils/core.test.ts
jmbtutor: Same as above.

packages/@sfx/core/test/unit/common/utils/core.test.ts
jmbtutor: Same as above.

packages/@sfx/core/test/unit/common/utils/core.test.ts
jmbtutor: Same as above.

packages/@sfx/core/test/unit/common/utils/core.test.ts
jmbtutor: Same as above.

packages/@sfx/core/test/unit/common/index.test.ts
```diff
@@ -0,0 +1,33 @@
...
```
DanielMoniz-GBI: Yeah, I think this is a good use case for comments. This feels very noisy, and I found the variable names confused me more than anything.

All of that said, we're testing an index file here. The actual value in the tests feels limited to me, so I don't want to spend too much time on it.

packages/@sfx/core/test/unit/common/core.test.ts
jmbtutor: I tried using TypeScript's `baseUrl` and `paths` compiler options, which enables [path mapping](https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping), but it looks like [`tsc` does not output resolved paths](https://github.com/microsoft/TypeScript/issues/26722).

There are two packages that look like they may solve that issue (namely [tscpaths](https://github.com/joonhocho/tscpaths) and [ts-transformer-inputs](https://github.com/grrowl/ts-transformer-imports)), but they don't seem to be widely used, and one of them explicitly says that it's not ready for production. The oft-recommended [module-alias](https://github.com/ilearnio/module-alias) is not suitable for us since this is a library and not an application. Doing it through Webpack is not an option since we don't use Webpack to build the package, and we don't want a single bundle for the whole package.

As much as I would like path aliasing, I don't think we can have it under our current circumstances unless we decide that the risk of using either tscpaths or ts-transformer-inputs is worth it.

packages/@sfx/core/test/unit/common/index.test.ts
jmbtutor: If I add a helper function to assert that two interfaces are equal, I think it'll make the test much clearer.

packages/@sfx/core/README.md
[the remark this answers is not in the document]
jmbtutor: I realized that if we do that, we would have no way to differentiate between a plugin and the options, especially if the options are optional (which it should be). In addition, I think the common case will be to register a number of plugins at the start, so an array is reasonable. I think I'll keep the array for those reasons.

packages/@sfx/core/src/plugin.ts
jmbtutor: Having the `PluginRegistry` makes it more explicit that a plugin registry is a dictionary, and allows us to change the type in the future if we so decide. It also provides the assertion that the keys can indeed by anything. Using the generic type `object` doesn't convey the intentions and is less flexible.

Having to specify `string` is required as a side effect of the syntax. This is what I meant by "unfortunately, yes".

packages/@sfx/core/test/unit/common/utils/core.test.ts
jmbtutor: Acknowledged. I'll keep it in the spirit of having more consistent-looking tests across the project and the org.

PR #10, opened 2019-07-16
SFX-160: Complete first pass of Sayt Component Driver Plugin

geoffcoutts:
Plugin serves as a link between our Sayt data source plugin and the Autocomplete web component.


packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -0,0 +1,123 @@
...
+  sendSaytAPIRequest(saytDataQuery: SaytDataPayload): Promise<string[]> {
+    const { query, ...config } = saytDataQuery;
+    return this.core.sayt.autocomplete(query, config, this.autocompleteCallback);
+  }
+
+  /**
+   * Callback for the Sayt client to transform the search response into
+   * relevant information.
+   *
+   * @param x Sayt client requires a placeholder attribute as the first argument
```
jmbtutor: This is not a placeholder parameter, this is for errors. We should checking this argument to make sure that there are no errors.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -0,0 +1,123 @@
...
+   * for any callback passed to it.
+   * @param response An array search term strings.
+   *
+   * @returns An array of search term strings.
+   */
+  autocompleteCallback(x: undefined, response: any): string[] {
```
jmbtutor:
```suggestion
  autocompleteCallback(error: Error, response: any): string[] {
```

It's fine to specify the parameter even if we don't use it, but in this case, we should probably be checking it.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  /**
+   * Callback for the Sayt data request event listener.
+   * Dispatches an event with the response from the sayt data plugin.
+   *
+   * @param saytDataQuery Request object received from the event listener.
+   */
+  fetchSaytData(saytDataQuery: SaytDataPayload): void {
+    const response = this.sendSaytAPIRequest(saytDataQuery);
+    response.then((data: any) => {
```
jmbtutor: I don't think this needs to be typed. TypeScript can infer the type, and if it can't, promises default to `any` anyway.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -0,0 +1,123 @@
...
+  /**
+   * Splits the request payload into the query string and a config object.
+   * Sends a search request to the Sayt data plugin with a callback to reformat.
+   *
+   * @param saytDataQuery Request object received from the event listener.
+   * @returns A promise from the Sayt API that has been reformatted
+   * with the passed callback.
+   */
+  sendSaytAPIRequest(saytDataQuery: SaytDataPayload): Promise<string[]> {
+    const { query, ...config } = saytDataQuery;
```
jmbtutor: You can fold this into the function signature, but don't do it if it looks less readable.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  /**
+   * Callback for the Sayt client to transform the search response into
+   * relevant information.
+   *
+   * @param x Sayt client requires a placeholder attribute as the first argument
+   * for any callback passed to it.
+   * @param response An array search term strings.
+   *
```
jmbtutor: Remove unnecessary line.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -0,0 +1,123 @@
...
+    response.then((data: any) => {
+      this.core[this.eventsPluginName].dispatchEvent(this.saytResponseEvent, data);
+    })
+    .catch((e) => {
+      this.core[this.eventsPluginName].dispatchEvent(this.saytErrorEvent, e);
+    });
+  }
+
+  /**
+   * Splits the request payload into the query string and a config object.
```
jmbtutor: The first sentence is the most important one and should stand alone as a short description of the function. The rest can follow as supplementary information.

Also, this sentence sounds like it probably doesn't need to be added because it's more of an implementation detail. Your doc comment should describe, without getting into implementation details, what you would expect the function to do when you call it. Something like "Sends a request to the SAYT API with the given query." would be suitable.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -0,0 +1,123 @@
...
+  autocompleteCallback(x: undefined, response: any): string[] {
+    return response.result.searchTerms.reduce((acc: string[], term: any) => {
+      if (term.value) acc.push(term.value);
+      return acc;
+    }, []);
+  }
+}
+
+/**
+ * Interface from the Sayt client for how the search config should be shaped.
```
jmbtutor:
```suggestion
 * The type of the sayt data event payload.
```

packages/@sfx/sayt-driver-plugin/test/unit/browser/sayt-driver-plugin.test.ts
```diff
@@ -0,0 +1,193 @@
+import { expect, spy, stub } from '../../utils';
+import { SaytDriverPlugin } from '../../../src/index';
+import { DomEventsPlugin } from '../../../../dom-events-plugin/src/index';
```
jmbtutor: This should import from the npm package.

```suggestion
import { DomEventsPlugin } from '@sfx/dom-events-plugin';
```

packages/@sfx/sayt-driver-plugin/test/unit/browser/sayt-driver-plugin.test.ts
```diff
@@ -0,0 +1,193 @@
...
+import { SaytPlugin } from '../../../../sayt-plugin/src';
+
+describe('Sayt Driver Plugin', () => {
+  let Driver: any;
+
+  beforeEach(() => {
+    Driver = new SaytDriverPlugin();
+    Driver.register({
+      'dom_events': new DomEventsPlugin().register({}),
+      'sayt': new SaytPlugin({
```
jmbtutor: For these two, instead of creating full plugins, it would be better to create stub instances. That way, you don't have to rely on either plugin actually working.

https://sinonjs.org/releases/latest/stubs/#var-stub--sinonstubobj

Alternatively—and this may be the better option—don't bother importing these and simply create objects that have the required methods.

packages/@sfx/sayt-driver-plugin/test/unit/browser/sayt-driver-plugin.test.ts
```diff
@@ -0,0 +1,193 @@
...
+    it('should specify a plugin name of sayt-driver', () => {
+      expect(Driver.metadata.name).to.equal('sayt_driver');
+    });
+
+    it('should have two dependencies: "dom_events" and "sayt"', () => {
+      expect(Driver.metadata.depends).to.deep.equal(['dom_events', 'sayt']);
+    });
+  });
+
+  describe('.register()', () => {
```
jmbtutor: No need for the `.`. This applies to the others as well.

```suggestion
  describe('register()', () => {
```

packages/@sfx/sayt-driver-plugin/test/unit/browser/sayt-driver-plugin.test.ts
```diff
@@ -0,0 +1,193 @@
...
+      'dom_events': new DomEventsPlugin().register({}),
+      'sayt': new SaytPlugin({
+        subdomain: 'macystest',
+        https: true,
+      }).register(),
+    });
+  });
+
+  describe('get metadata()', () => {
+    it('should specify a plugin name of sayt-driver', () => {
```
jmbtutor:
```suggestion
    it('should specify a plugin name of sayt_driver', () => {
```

packages/@sfx/sayt-driver-plugin/test/unit/browser/sayt-driver-plugin.test.ts
```diff
@@ -0,0 +1,193 @@
+import { expect, spy, stub } from '../../utils';
+import { SaytDriverPlugin } from '../../../src/index';
+import { DomEventsPlugin } from '../../../../dom-events-plugin/src/index';
+import { SaytPlugin } from '../../../../sayt-plugin/src';
+
+describe('Sayt Driver Plugin', () => {
+  let Driver: any;
```
jmbtutor: Use a lowercase `driver`. This is not a class. This can also be implicitly typed, so you don't need to type it.
```suggestion
  let driver;
```

packages/@sfx/sayt-driver-plugin/test/unit/browser/sayt-driver-plugin.test.ts
```diff
@@ -0,0 +1,193 @@
...
+    });
+  });
+
+  describe('get metadata()', () => {
+    it('should specify a plugin name of sayt-driver', () => {
+      expect(Driver.metadata.name).to.equal('sayt_driver');
+    });
+
+    it('should have two dependencies: "dom_events" and "sayt"', () => {
+      expect(Driver.metadata.depends).to.deep.equal(['dom_events', 'sayt']);
```
jmbtutor: We don't care that the array is in a certain order, so it would be better to assert that they include the expected items instead.

```suggestion
      expect(Driver.metadata.depends).to.have.members(['dom_events', 'sayt']);
```

packages/@sfx/sayt-driver-plugin/test/unit/browser/sayt-driver-plugin.test.ts
```diff
@@ -0,0 +1,193 @@
...
+import { SaytDriverPlugin } from '../../../src/index';
+import { DomEventsPlugin } from '../../../../dom-events-plugin/src/index';
+import { SaytPlugin } from '../../../../sayt-plugin/src';
+
+describe('Sayt Driver Plugin', () => {
+  let Driver: any;
+
+  beforeEach(() => {
+    Driver = new SaytDriverPlugin();
+    Driver.register({
```
jmbtutor: This should not be relying on `register` actually working. Each test should instead inject an object into `driver.core` that contains the expected plugins by setting the `core` property directly.

packages/@sfx/sayt-driver-plugin/test/unit/browser/sayt-driver-plugin.test.ts
```diff
@@ -0,0 +1,193 @@
...
+      expect(Driver.metadata.name).to.equal('sayt_driver');
+    });
+
+    it('should have two dependencies: "dom_events" and "sayt"', () => {
+      expect(Driver.metadata.depends).to.deep.equal(['dom_events', 'sayt']);
+    });
+  });
+
+  describe('.register()', () => {
+    beforeEach(() => {
```
jmbtutor: If you set `driver.core` in the individual tests instead of the `beforeEach` as above, this no longer becomes necessary.

packages/@sfx/sayt-driver-plugin/test/unit/browser/sayt-driver-plugin.test.ts
```diff
@@ -0,0 +1,193 @@
...
+      Driver = {};
+    })
+
+    it('should keep a reference to all registered plugins', () => {
+      const Plugins = {
+        a: () => 'a',
+        b: () => true,
+      }
+      Driver = new SaytDriverPlugin();
```
jmbtutor: Ditto with the `driver.core` comment.

packages/@sfx/sayt-driver-plugin/test/unit/browser/sayt-driver-plugin.test.ts
```diff
@@ -0,0 +1,193 @@
...
+    });
+  });
+
+  describe('.register()', () => {
+    beforeEach(() => {
+      Driver = {};
+    })
+
+    it('should keep a reference to all registered plugins', () => {
+      const Plugins = {
```
jmbtutor: This is simply an object, so this should be lowercase.

```suggestion
      const plugins = {
```

packages/@sfx/sayt-driver-plugin/test/unit/browser/sayt-driver-plugin.test.ts
```diff
@@ -0,0 +1,193 @@
...
+        a: () => 'a',
+        b: () => true,
+      }
```
jmbtutor:
```suggestion
      };
```

packages/@sfx/sayt-driver-plugin/test/unit/browser/sayt-driver-plugin.test.ts
```diff
@@ -0,0 +1,193 @@
...
+      Driver = new SaytDriverPlugin();
+
+      expect(Driver.core).to.be.undefined;
+      Driver.register(Plugins);
+
+      expect(Driver.core).to.deep.equal(Plugins);
```
jmbtutor: You actually want it to be the exact same reference, otherwise it won't get any updates to the registry, so take out `.deep`.
```suggestion
      expect(Driver.core).to.equal(Plugins);
```

packages/@sfx/sayt-driver-plugin/test/unit/browser/sayt-driver-plugin.test.ts
```diff
@@ -0,0 +1,193 @@
...
+    })
+
+    it('should keep a reference to all registered plugins', () => {
+      const Plugins = {
+        a: () => 'a',
+        b: () => true,
+      }
+      Driver = new SaytDriverPlugin();
+
+      expect(Driver.core).to.be.undefined;
```
jmbtutor: Ditto with the `driver.core` comment.

packages/@sfx/sayt-driver-plugin/test/unit/browser/sayt-driver-plugin.test.ts
```diff
@@ -0,0 +1,193 @@
...
+      Driver.register(Plugins);
+
+      expect(Driver.core).to.deep.equal(Plugins);
+    });
+  });
+
+  describe('.ready()', () => {
+    let registerListener: any;
```
jmbtutor: No type annotation necessary.

packages/@sfx/sayt-driver-plugin/test/unit/browser/sayt-driver-plugin.test.ts
```diff
@@ -0,0 +1,193 @@
...
+
+    beforeEach(() => {
+      registerListener = spy(Driver.core['dom_events'], 'registerListener');
```
jmbtutor: This should be stubbing the function because you don't actually want to register a listener. But also, ditto with the `driver.core` comment.

packages/@sfx/sayt-driver-plugin/test/unit/browser/sayt-driver-plugin.test.ts
```diff
@@ -0,0 +1,193 @@
...
+    });
+
+    it('should register an event listener for receiving sayt API requests', () => {
+      Driver.ready();
+
+      registerListener.getCall(0).calledWith(Driver.saytDataEvent, Driver.fetchSaytData);
```
jmbtutor:
```suggestion
      expect(registerListener).to.be.calledWith(Driver.saytDataEvent, Driver.fetchSaytData);
```

packages/@sfx/sayt-driver-plugin/test/unit/browser/sayt-driver-plugin.test.ts
```diff
@@ -0,0 +1,193 @@
...
+    });
+  });
+
+  describe('.unregister()', () => {
+    let unregisterListener;
+
+    beforeEach(() => {
+      unregisterListener = spy(Driver.core['dom_events'], 'unregisterListener');
```
jmbtutor: Same comment as with the above.

packages/@sfx/sayt-driver-plugin/test/unit/browser/sayt-driver-plugin.test.ts
```diff
@@ -0,0 +1,193 @@
...
+    });
+
+    it('should unregister the sayt event listener', () => {
+      Driver.unregister();
+
+      expect(unregisterListener).to.have.been.calledWith(Driver.saytDataEvent, Driver.fetchSaytData);
+    });
+  });
+
+  describe('.autocompleteCallback', () => {
```
jmbtutor: Keep it consistent. Add the `()`.

```suggestion
  describe('autocompleteCallback()', () => {
```

packages/@sfx/sayt-driver-plugin/test/unit/browser/sayt-driver-plugin.test.ts
```diff
@@ -0,0 +1,193 @@
...
+        result: {
+          searchTerms: [
+            { value: 'a' , notValue: 'z' },
+            { value: 'b' },
+            { value: 'c' },
+          ],
+        },
+      };
+
+      const cbReturn = Driver.autocompleteCallback(undefined, response);
```
jmbtutor: Name this something more descriptive, like "terms".

packages/@sfx/sayt-driver-plugin/test/unit/browser/sayt-driver-plugin.test.ts
```diff
@@ -0,0 +1,193 @@
...
+
+      expect(cbReturn).to.deep.equal([]);
+    });
+  });
+
+  describe('.sendSaytAPIRequest()', () => {
+    const saytDataPayload = {
```
jmbtutor: This should be set in the `beforeEach` to maintain test isolation.

packages/@sfx/sayt-driver-plugin/test/unit/browser/sayt-driver-plugin.test.ts
```diff
@@ -0,0 +1,193 @@
...
+      collection: 'backup',
+    };
+    let autocompleteCallback;
+
+    beforeEach(() => {
+      autocompleteCallback = stub(Driver, 'autocompleteCallback');
+    });
+
+    it('should make a search call through the sayt client', () => {
+      const autocomplete = spy(Driver.core.sayt, 'autocomplete');
```
jmbtutor: Use a stub. Ditto `driver.core` comment. If you need the stub to return a promise, you can look at [`stub.resolves`](https://sinonjs.org/releases/latest/stubs/#stubresolvesvalue).

packages/@sfx/sayt-driver-plugin/test/unit/browser/sayt-driver-plugin.test.ts
```diff
@@ -0,0 +1,193 @@
...
+      const callbackReturn = autocompleteCallback();
+
+      const returnValue = Driver.sendSaytAPIRequest(saytDataPayload);
+
+      expect(returnValue).to.eventually.equal(callbackReturn);
+    });
+  });
+
+  describe('.fetchSaytData()', () => {
+    const query = {
```
jmbtutor: This should be set in the beforeEach to maintain test isolation.

packages/@sfx/sayt-driver-plugin/test/unit/browser/sayt-driver-plugin.test.ts
```diff
@@ -0,0 +1,193 @@
...
+      query: 'shirt',
+    };
+    let dispatchEvent;
+    let sendSaytAPIRequest;
+
+    beforeEach(() => {
+      sendSaytAPIRequest = stub(Driver, 'sendSaytAPIRequest').callThrough();
```
jmbtutor: If you're going to `callThrough()`, this may as well be a spy instead. But in either case, it _shouldn't_ call through.

packages/@sfx/sayt-driver-plugin/test/unit/browser/sayt-driver-plugin.test.ts
```diff
@@ -0,0 +1,193 @@
...
+      dispatchEvent = spy(Driver.core['dom_events'], 'dispatchEvent');
```
jmbtutor: This should be a stub.

packages/@sfx/sayt-driver-plugin/test/unit/browser/sayt-driver-plugin.test.ts
```diff
@@ -0,0 +1,193 @@
...
+      expect(sendSaytAPIRequest).to.be.calledWith(query);
+    });
+
+    it('should dispatch the response through the events plugin', () => {
+      const response = { a: 'b' };
+      sendSaytAPIRequest.returns(response);
+
+      Driver.fetchSaytData(query);
+
+      expect(Promise.resolve(dispatchEvent))
```
jmbtutor: Assertions using chai-as-promised should be returned so that mocha can wait on it.

packages/@sfx/sayt-driver-plugin/test/unit/browser/sayt-driver-plugin.test.ts
```diff
@@ -0,0 +1,193 @@
...
+        .to.be.eventually.calledOnceWith(Driver.saytResponseEvent, response);
+    });
+
+    it('should send an error in an event if the API request fails', () => {
+      sendSaytAPIRequest.rejects('test error');
+
+      Driver.fetchSaytData(query);
+
+      expect(Promise.resolve(dispatchEvent))
```
jmbtutor: Assertions using chai-as-promised should be returned so that mocha can wait on it.

packages/@sfx/sayt-driver-plugin/test/unit/browser/sayt-driver-plugin.test.ts
```diff
@@ -0,0 +1,193 @@
...
+      );
+    });
+
+    it('should return the result of the Sayt API callback', () => {
+      autocompleteCallback.returns(['a', 'b']);
+      const callbackReturn = autocompleteCallback();
+
+      const returnValue = Driver.sendSaytAPIRequest(saytDataPayload);
+
+      expect(returnValue).to.eventually.equal(callbackReturn);
```
jmbtutor: Assertions using chai-as-promised should be returned so that mocha can wait on it.

packages/@sfx/sayt-driver-plugin/test/unit/browser/sayt-driver-plugin.test.ts
jmbtutor: Again, if you need the stub to return a promise, you can look at [`stub.resolves`](https://sinonjs.org/releases/latest/stubs/#stubresolvesvalue) and [`stub.rejects`](https://sinonjs.org/releases/latest/stubs/#stubrejects).

packages/@sfx/sayt-driver-plugin/test/unit/common/index.test.ts
```diff
@@ -0,0 +1,9 @@
+import { expect } from '../../utils';
+import { SaytDriverPlugin as SaytDriverPluginExport} from '../../../src';
```
jmbtutor:
```suggestion
import { SaytDriverPlugin as SaytDriverPluginExport } from '../../../src';
```

packages/@sfx/sayt-driver-plugin/test/unit/common/index.test.ts
```diff
@@ -0,0 +1,9 @@
...
+import SaytDriverPlugin from '../../../src/sayt-driver-plugin';
+
+describe('Entry point', () => {
+  it('should export EventsBrowserPlugin ', () => {
+    expect(SaytDriverPluginExport).to.equal(SaytDriverPlugin);
+  });
+});
```
jmbtutor: I'm going to ignore the lack of interface exports for now because we decided to defer that decision.

packages/@sfx/sayt-plugin/test/unit/common/sayt-plugin.test.ts
```diff
@@ -39,5 +39,10 @@ describe('SaytPlugin', () => {
 
       expect(registerReturnValue).to.equal(saytInstance);
     });
+
+    it('should expose an autocomplete method', () => {
```
jmbtutor: I don't think we need to test this.

presets/plugins.ts
```diff
@@ -1 +1,2 @@
 export * from '@sfx/dom-events-plugin';
+export * from '@sfx/sayt-driver-plugin/src';
```
jmbtutor: `/src` should not be necessary.

packages/@sfx/sayt-driver-plugin/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "@sfx/sayt-driver-plugin",
+  "version": "0.0.0",
+  "description": "",
+  "main": "dist/index.js",
+  "module": "src/index.ts",
+  "types": "dist/index.d.ts",
```
jmbtutor: Make sure to include `esnext`. Check the other packages for the correct order to keep it consistent.

packages/@sfx/sayt-driver-plugin/test/unit/browser/sayt-driver-plugin.test.ts
DanielMoniz-GBI: @geoffcoutts, you could try skipping this test to see if the assertion is actually included in the final total.

packages/@sfx/sayt-plugin/test/unit/common/sayt-plugin.test.ts
DanielMoniz-GBI: We added this test because we were unsure of whether or not the SAYT client was actually being exposed correctly. We test above that it exposes its `.sayt` property, but not that a useful SAYT client is actually exposed.

jmbtutor: Sure, but the constructor test already tests that we're creating an instance of `Sayt`, which has an `autocomplete` method (not explicitly stated anywhere in the test, but it's implicit because of the `Sayt` interface). That test and the test above taken together imply that we have an `autocomplete` method. That said, one could make the argument that that's an implementation detail and that we _should_ be testing this, but if so, I would argue that we should be testing that the `autocomplete` method to make sure that it's useful, not simply that we expose one.

packages/@sfx/sayt-driver-plugin/test/unit/browser/sayt-driver-plugin.test.ts
jmbtutor: Skipping the test wouldn't have an effect. This test passes because no exceptions are thrown during its runtime. A better way to check is to resolve to something that's explicitly not the expected return value and see if it fails.

packages/@sfx/sayt-plugin/test/unit/common/sayt-plugin.test.ts
DanielMoniz-GBI: That's true. I'd suggest we have an integration test for this kind of thing, given that the driver's job is to pull a few different packages together. Next sprint we'll likely be pulling in some testing tickets.

packages/@sfx/sayt-driver-plugin/test/unit/browser/sayt-driver-plugin.test.ts
geoffcoutts: I've changed the injected events plugin for the tests to have methods that return `null` instead of setting a listener now.

packages/@sfx/sayt-driver-plugin/test/unit/common/sayt-driver-plugin.test.ts
```diff
@@ -0,0 +1,272 @@
+import { expect, spy, stub } from '../../utils';
+import { SaytDriverPlugin } from '../../../src/index';
+
+describe('Sayt Driver Plugin', () => {
+  let driver;
+  let dom_events;
```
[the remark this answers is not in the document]
geoffcoutts: `dom_events` is the plugin name that it has a dependency on.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -0,0 +1,146 @@
...
+    return terms.map((term) => {
+      return { label: term.value };
+    }).filter((item) => item.label);
+  }
+}
+
+/**
+ * The type of the sayt autocomplete request event payload.
+ */
+export interface QueryTimeAutocompleteConfigWithQuery extends QueryTimeAutocompleteConfig {
```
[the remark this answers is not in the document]
geoffcoutts: I can drop the `QueryTime` if you want?

geoffcoutts:
> What is the purpose of the empty `.gitignore` file? I noticed it exists in other plugins on our `master` branch right now as well.

It is to maintain the empty folders for a consistent folder structure across the plugins. It probably should have been more aptly named `.gitkeep`, but i'm following the convention that has been started.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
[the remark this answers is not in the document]
DanielMoniz-GBI: Yeah, let's just do that. Presumably if a query is supplied, that implies that it's query-time.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -105,15 +106,48 @@ export default class SaytDriverPlugin implements Plugin {
...
+        return { label: term.value };
+    });
   }
 }
 
 /**
  * The type of the sayt autocomplete request event payload.
  */
-export interface QueryTimeAutocompleteConfigWithQuery extends QueryTimeAutocompleteConfig {
+export interface AutocompleteConfigWithQuery extends QueryTimeAutocompleteConfig {
```
jmbtutor: I think `AutocompleteConfig` is sufficient here, unless we have one without `query`. Having `WithQuery` communicates to me that we have one that doesn't.

packages/@sfx/sayt-driver-plugin/CHANGELOG.md
```diff
@@ -0,0 +1,10 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+### Added
+- SFX-160: Added the `SaytDriverPlugin` module.
+  - The SaytDriverPlugin module acts as a mediator between the logic and view layers for prebuilt modules.
```
jmbtutor:
```suggestion
  - The `SaytDriverPlugin` module acts as a mediator between the Logic and View layers for prebuilt modules.
```

PR #12, opened 2019-07-23
SFX-160: Fix event handler

geoffcoutts:
Fix the event handler to properly accept and event as its argument and then grab the event payload from inside.


packages/@sfx/sayt-driver-plugin/test/unit/common/sayt-driver-plugin.test.ts
```diff
@@ -241,15 +243,15 @@ describe('Sayt Driver Plugin', () => {
     it('should get a response from Sayt client request method', () => {
       sendSaytApiRequest.resolves(response);
 
-      driver.fetchSaytData(query);
+      driver.fetchSaytData(fetchEvent);
 
-      expect(sendSaytApiRequest).to.be.calledWith(query);
+      expect(sendSaytApiRequest).to.be.calledWith(fetchEvent.detail);
```
jmbtutor: I would pull out `detail` into a `const` and use it here instead of drilling into `fetchEvent`. It gives you more control.

PR #16, opened 2019-07-25
SFX-159: Add Search Driver plugin

jmbtutor:
The search driver is responsible for listening for events, making search requests when those events are received, and dispatching the results through other events.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -106,6 +106,9 @@ jobs:
       - restore_cache:
           key: build_cache-v{{ .Environment.CI_CACHE_VERSION }}-{{ .Branch }}-{{ .Revision }}-{{ checksum "../workflow-start" }}-build_sayt_driver_plugin-
       - run: *extract_build_archive
+      - restore_cache:
+          key: build_cache-v{{ .Environment.CI_CACHE_VERSION }}-{{ .Branch }}-{{ .Revision }}-{{ checksum "../workflow-start" }}-build_search_driver_plugin-
+      - run: *extract_build_archive
       - run: *archive_build
       - save_cache:
           key: build_cache-v{{ .Environment.CI_CACHE_VERSION }}-{{ .Branch }}-{{ .Revision }}-{{ checksum "../workflow-start" }}-all-{{ epoch }}
@@ -197,10 +200,15 @@ workflows:
           name: build_sayt_driver_plugin
           requires:
             - collect_independent_builds
+      - build:
+          name: build_search_driver_plugin
+          requires:
+            - collect_independent_builds
       - collect_builds:
           name: build_final
           requires:
             - build_sayt_driver_plugin
+            - build_search_driver_plugin
 
       - unit_test:
           name: unit_test_core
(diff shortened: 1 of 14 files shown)


packages/@sfx/search-driver-plugin/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "@sfx/search-driver-plugin",
+  "version": "0.0.0",
+  "description": "",
+  "main": "dist/index.js",
+  "module": "src/index.ts",
+  "scripts": {
```
ssbelle: do we need an esnext line here @jmbtutor ?

packages/@sfx/search-driver-plugin/src/search-driver-plugin.ts
```diff
@@ -0,0 +1,88 @@
...
+        this.core[this.eventsPluginName].dispatchEvent(SEARCH_ERROR_EVENT, e);
+      });
+  }
+}
+
+/**
+ * The type of the search request event payload. The payload is the
+ * search term.
+ */
+export type SearchRequestPayload = string;
```
DanielMoniz-GBI: Ahh, this is all you meant before. Got it.

geoffcoutts:
@jmbtutor Currently getting two typescript build errors when building locally. It looks like this needs to be added to the CircleCI config too.

packages/@sfx/search-driver-plugin/src/search-driver-plugin.ts
```diff
@@ -0,0 +1,88 @@
...
+   * through an event. The result is emitted in a
+   * [[SEARCH_RESPONSE_EVENT]] event. If the search fails for any
+   * reason, a [[SEARCH_ERROR_EVENT]] is dispatched with the error.
+   *
+   * @param event the event whose payload is the search term.
+   */
+  fetchSearchData(event: CustomEvent<SearchRequestPayload>): void {
+    const searchTerm = event.detail;
+    this.core.search.search(searchTerm)
+      .then((results) => {
```
JoeyDeol: The `results` parameter has an implicit type of any, it needs to be set explicitly if it is to remain as any. Otherwise is there a way for us to give it a specific type?

packages/@sfx/search-driver-plugin/src/search-driver-plugin.ts
```diff
@@ -0,0 +1,88 @@
...
+        this.core[this.eventsPluginName].dispatchEvent(SEARCH_RESPONSE_EVENT, results);
+      })
+      .catch((e) => {
```
JoeyDeol: Same as above but for the `e` parameter

.circleci/config.yml
```diff
@@ -233,6 +241,9 @@ workflows:
       - unit_test_browser:
           name: unit_test_browser_sayt_driver_plugin
           <<: *requires_build_final
+      - unit_test_browser:
```
geoffcoutts: Does this need to be added to the `unt_test` jobs section as well?

jmbtutor: Yes

PR #18, opened 2019-08-02
Integrate Logic and View layers

ssbelle:
- Added a bind to fix sayt functionality 
- Created file to export core and all plugins
- Added all plugins to the export plugins file
- Added a line in webpack to bundle this temp bundle of core and all plugins as 'sfx-all-plugins-bundle.js'

### DIFF
--- a/presets/plugins.ts
+++ b/presets/plugins.ts
@@ -1,5 +1,5 @@
 export * from '@sfx/dom-events-plugin';
 export * from '@sfx/sayt-driver-plugin';
 export * from '@sfx/sayt-plugin';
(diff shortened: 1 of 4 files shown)


presets/plugins.ts
```diff
@@ -1,2 +1,5 @@
 export * from '@sfx/dom-events-plugin';
 export * from '@sfx/sayt-driver-plugin';
+export * from '@sfx/sayt-plugin';
+export * from '@sfx/search-plugin';
+export * from '@sfx/search-driver-plugin';
```
JoeyDeol: Should we order these alphabetically?

PR #20, opened 2019-08-15
SFX-295: Modify saytDriverPlugin to pass searchbox ID

ssbelle:


packages/@sfx/sayt-driver-plugin/CHANGELOG.md
```diff
@@ -8,3 +8,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - SFX-160: Added the `SaytDriverPlugin` module.
   - The `SaytDriverPlugin` module acts as a mediator between the Logic and View layers for prebuilt modules.
+
+- SFX-295: Added the origin searchbox ID to the `saytResponseEvent`
```
jmbtutor: Since this is still unreleased, I would add this as a sub-point under SFX-160. Like so:

```markdown
- SFX-160: Added the `SaytDriverPlugin` module.
  - The `SaytDriverPlugin` module acts as a mediator between the Logic and View layers for prebuilt modules.
  - SFX-295: Added the associated searchbox ID to response events.
```

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -75,15 +75,18 @@ export default class SaytDriverPlugin implements Plugin {
   }
 
   /**
-   * Dispatches an event with the response from the sayt data plugin.
+   * Dispatches an event with the response from the sayt data plugin and the origin searchbox id.
```
jmbtutor: I'd say "associated" instead of "origin". "ID" should be in capital letters.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -75,15 +75,18 @@ export default class SaytDriverPlugin implements Plugin {
...
    * Callback for the Sayt data request event listener.
    *
    * @param event Event that contains the Sayt API request payload.
    */
   fetchSaytData(event: CustomEvent): void {
+    const searchbox = { searchbox: event.detail.searchbox };
```
JoeyDeol: Please refer to my comment following this one: if you implement the feedback below then you won't need this variable. Since you're only using this in one spot I don't think you need to assign it to a variable.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -75,15 +75,18 @@ export default class SaytDriverPlugin implements Plugin {
...
     this.sendSaytApiRequest(event.detail)
```
geoffcoutts: This will now send the searchbox id to the Sayt API. We should try and prevent sending any pieces of information that are not expected. I would suggest changes to deconstruct the original `event.detail` in this function and only pass the required pieces to `sendSaytAPIRequest()`.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -75,15 +75,18 @@ export default class SaytDriverPlugin implements Plugin {
...
       .then((data) => {
-        this.core[this.eventsPluginName].dispatchEvent(this.saytResponseEvent, data);
+        const dataObject = Object.assign(data, searchbox);
```
geoffcoutts: Is there a more descriptive name that we can use?

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
ssbelle: @geoffcoutts I don't think Im sending the searchbox id to the API, I thought what I was doing was getting the response from the api, then() adding the searchbox to the response and sending searchbox with response in the event on line 89

geoffcoutts: The ID is part of `event.detail` which is getting passed to `sendSaytAPIRequest()`. This deconstructs the event into `query` and `config`, which will now contain the ID, and sends both pieces to the API.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -75,15 +75,16 @@ export default class SaytDriverPlugin implements Plugin {
...
    * @param event Event that contains the Sayt API request payload.
    */
   fetchSaytData(event: CustomEvent): void {
-    this.sendSaytApiRequest(event.detail)
-      .then((data) => {
-        this.core[this.eventsPluginName].dispatchEvent(this.saytResponseEvent, data);
+    const { query } = event.detail;
+    this.sendSaytApiRequest({ query })
+      .then((results) => {
+        this.core[this.eventsPluginName].dispatchEvent(this.saytResponseEvent, { results, searchbox: event.detail.searchbox });
```
JoeyDeol: With the latest update, you can just pull out `searchbox` from `event.detail` in the same place you are pulling out `query`. That way, you just have to refer to the `searchbox` variable in the new object you are dispatching with the event.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -75,15 +75,16 @@ export default class SaytDriverPlugin implements Plugin {
...
+   * Dispatches an event with the response from the sayt data plugin and the associated searchbox ID.
    * Callback for the Sayt data request event listener.
    *
    * @param event Event that contains the Sayt API request payload.
    */
   fetchSaytData(event: CustomEvent): void {
-    this.sendSaytApiRequest(event.detail)
-      .then((data) => {
-        this.core[this.eventsPluginName].dispatchEvent(this.saytResponseEvent, data);
+    const { query, searchbox } = event.detail;
```
geoffcoutts: This was more along what I had in mind as well still need to pass in the config object. 

```suggestion
    const { query, searchbox, ...config } = event.detail;
    this.sendSaytApiRequest(query, config)
```

geoffcoutts: Actually, in talking with @jmbtutor, a better approach would be to not spread `config`. This makes it a little safer and avoids the issue with sending unexpected information to the Sayt API in the future. 

Will also require changing the signature for `sendSaytApiRequest() slightly`.

packages/@sfx/sayt-driver-plugin/test/unit/common/sayt-driver-plugin.test.ts
```diff
@@ -202,11 +202,18 @@ describe('Sayt Driver Plugin', () => {
     });
 
     it('should make a search call through the sayt client', () => {
-      driver.sendSaytApiRequest(saytDataPayload);
+      const payload = {
```
geoffcoutts: Should probably change the original variable setup in the `beforeEach()` step instead and then modify the two tests accordingly.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -136,6 +137,7 @@ export default class SaytDriverPlugin implements Plugin {
  */
 export interface AutocompleteRequestConfig extends QueryTimeAutocompleteConfig {
   query: string;
+  config: any;
```
geoffcoutts: `config` is the `QueryTimeAutocompleteConfig`.  This interface should be changed to no longer extend `QueryTimeAutocompleteConfig` and `config` should be of that type.

packages/@sfx/sayt-driver-plugin/test/unit/common/sayt-driver-plugin.test.ts
```diff
@@ -16,6 +19,22 @@ describe('Sayt Driver Plugin', () => {
       autocomplete: (query, config, callback) => null,
     };
     driver = new SaytDriverPlugin();
+    validConfig = {
+      config: {
```
geoffcoutts: This doesn't need to live inside of another object.

packages/@sfx/sayt-driver-plugin/test/unit/common/sayt-driver-plugin.test.ts
```diff
@@ -16,6 +19,22 @@ describe('Sayt Driver Plugin', () => {
...
+        collection: 'backup'
+      }
+    };
+    fetchEvent = {
+      detail: {
+        query: 'shirt',
```
geoffcoutts: This can just be a variable on its own since you are injecting it into `saytDataPayload` below. Then you can reference it through the payload.

packages/@sfx/sayt-driver-plugin/test/unit/common/sayt-driver-plugin.test.ts
```diff
@@ -222,40 +236,44 @@ describe('Sayt Driver Plugin', () => {
...
 
     it('should dispatch the response through the events plugin', () => {
-      sendSaytApiRequest.resolves(response);
+      sendSaytApiRequest.resolves(results);
 
       driver.fetchSaytData(fetchEvent);
 
       return expect(Promise.resolve(dispatchEvent))
-        .to.be.eventually.calledOnceWith(driver.saytResponseEvent, response);
+        .to.be.eventually.calledWith(driver.saytResponseEvent, { results, searchbox });
```
geoffcoutts: Should change this back to `calledOnceWith`

packages/@sfx/sayt-driver-plugin/test/unit/common/sayt-driver-plugin.test.ts
```diff
@@ -222,40 +236,44 @@ describe('Sayt Driver Plugin', () => {
...
-      };
-      response = { a: 'b' };
+      results =  { a: "b" };
       sendSaytApiRequest = stub(driver, 'sendSaytApiRequest');
     });
 
-    it('should get a response from Sayt client request method', () => {
-      const saytPayload  = fetchEvent.detail;
-      sendSaytApiRequest.resolves(response);
+    it('should call sendSaytApiRequest with query from event and undefined config', () => {
```
geoffcoutts:
```suggestion
    it('should call sendSaytApiRequest with query from event and an undefined config if it is empty', () => {
```

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -136,6 +137,7 @@ export default class SaytDriverPlugin implements Plugin {
  */
 export interface AutocompleteRequestConfig extends QueryTimeAutocompleteConfig {
```
jmbtutor: This interface needs a `searchbox?: string` property.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -75,15 +75,16 @@ export default class SaytDriverPlugin implements Plugin {
   }
 
   /**
-   * Dispatches an event with the response from the sayt data plugin.
+   * Dispatches an event with the response from the sayt data plugin and the associated searchbox ID.
    * Callback for the Sayt data request event listener.
    *
    * @param event Event that contains the Sayt API request payload.
    */
   fetchSaytData(event: CustomEvent): void {
```
jmbtutor: You can give the payload a type by passing in a type parameter.

```suggestion
  fetchSaytData(event: CustomEvent<AutocompleteRequestConfig>): void {
```

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
jmbtutor: For the record, these properties should have doc comments attached to them, but you don't need to do them here because we have another ticket with proper comments already.

packages/@sfx/sayt-driver-plugin/test/unit/common/sayt-driver-plugin.test.ts
```diff
@@ -16,6 +19,22 @@ describe('Sayt Driver Plugin', () => {
...
+      }
+    };
+    fetchEvent = {
+      detail: {
+        query: 'shirt',
+      }
+    };
+    saytDataPayload = {
+      detail: {
+        ...fetchEvent.detail,
```
jmbtutor: You should consider it a red flag when you have to drill down through an object that you've create to get a value because it's one less level of control that you have. Generally, that means that the value you want should instead be put into its own variable. That applies to the rest of the tests.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
jmbtutor: `config` should also be optional.

packages/@sfx/sayt-driver-plugin/test/unit/common/sayt-driver-plugin.test.ts
jmbtutor: I don't think this test is necessary. I think we just need to ensure that the function doesn't throw when `config` is not specified. I'm unsure if even that test is necessary, but it wouldn't hurt.

packages/@sfx/sayt-driver-plugin/test/unit/common/sayt-driver-plugin.test.ts
```diff
@@ -222,40 +236,44 @@ describe('Sayt Driver Plugin', () => {
...
+      results =  { a: "b" };
       sendSaytApiRequest = stub(driver, 'sendSaytApiRequest');
     });
 
-    it('should get a response from Sayt client request method', () => {
-      const saytPayload  = fetchEvent.detail;
-      sendSaytApiRequest.resolves(response);
+    it('should call sendSaytApiRequest with query from event and undefined config', () => {
+      const saytQueryPayload  = fetchEvent.detail;
+      const undefinedConfig = { config: undefined }
```
jmbtutor: Generally, instead of setting things to `undefined`, you often can just leave it out.

packages/@sfx/sayt-driver-plugin/test/unit/common/sayt-driver-plugin.test.ts
```diff
@@ -16,6 +20,23 @@ describe('Sayt Driver Plugin', () => {
       autocomplete: (query, config, callback) => null,
     };
     driver = new SaytDriverPlugin();
+    config = {
+      collection: 'backup'
+    };
+    query = {
```
DanielMoniz-GBI: Shouldn't this just be a string? The tests below seem to assert that `autocomplete()` will receive a query object, which doesn't seem to be true. From the following definition:
```
sendSaytApiRequest(query: string, config: QueryTimeAutocompleteConfig): Promise<string[]> {
    return this.core.sayt.autocomplete(query, config).then(this.autocompleteCallback);
  }
```
it appears that `this.core.sayt.autocomplete()` will receive a string as its first argument, as indicated in the parameters for `sendSaytApiRequest()`.

Ultimately it doesn't matter for the tests if we're mocking out that function, but I think we should at least try to be accurate in terms of the basic shape of the data we use to query.

packages/@sfx/sayt-driver-plugin/test/unit/common/sayt-driver-plugin.test.ts
```diff
@@ -16,6 +20,21 @@ describe('Sayt Driver Plugin', () => {
       autocomplete: (query, config, callback) => null,
     };
     driver = new SaytDriverPlugin();
+    config = {
+      collection: 'backup'
```
geoffcoutts:
```suggestion
      collection: 'backup',
```

packages/@sfx/sayt-driver-plugin/test/unit/common/sayt-driver-plugin.test.ts
```diff
@@ -2,9 +2,13 @@ import { expect, spy, stub } from '../../utils';
 import { SaytDriverPlugin } from '../../../src/index';
 
 describe('Sayt Driver Plugin', () => {
+  let config;
   let driver;
   let dom_events;
   let sayt;
+  let saytDataPayload;
+  let saytDataUndefinedConfig;
+  let query;
```
jmbtutor: The value of `query` is a primitive, so it's pointless to reassign it every time in `beforeEach` because there's no chance that it can be modified.

```suggestion
  const query = 'shirt';
```

packages/@sfx/sayt-driver-plugin/test/unit/common/sayt-driver-plugin.test.ts
```diff
@@ -16,6 +20,21 @@ describe('Sayt Driver Plugin', () => {
       autocomplete: (query, config, callback) => null,
     };
     driver = new SaytDriverPlugin();
+    config = {
+      collection: 'backup'
+    };
+    query = 'shirt';
```
jmbtutor: With the suggestion above about making `query` a `const`, you can remove this.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -97,7 +99,7 @@ export default class SaytDriverPlugin implements Plugin {
    * @returns A promise from the Sayt API that has been reformatted
    * with the passed callback.
    */
```
jmbtutor: This doc comment needs to be updated with the new parameters.

```
@param query The search term to send.
@param config Extra query-time configuration to customize the SAYT request.
```

ssbelle:
> Nice work!

Thanks for your help!

PR #23, opened 2019-08-26
SFX-333: Provide products for SAYT

DanielMoniz-GBI:


packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -33,19 +35,41 @@ export default class SaytDriverPlugin implements Plugin {
...
+   */
+  productResponseEvent: string = 'sfx::sayt_products_response';
+  /**
+   * Event name to listen for Sayt product errors.
+   */
+  productErrorEvent: string = 'sfx::sayt_products_error';
+  /**
+   * Provide default configuration for SAYT product searches.
+   */
+  defaultSearchConfig: object = {
```
jmbtutor: Do we have an interface for this somewhere?

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -120,6 +174,55 @@ export default class SaytDriverPlugin implements Plugin {
...
+  }
+
+  /**
+   * Parses a given product record for valid data.
+   *
+   * @param record An object containing the product record data.
+   * @returns An object containing relevant product data,
+   * or undefined if record is invalid.
+   */
+  parseRecord(record: Record): any | undefined {
```
jmbtutor: `any | undefined` is redundant. Since you know the shape of the data you're returning, `any` should be a concrete interface.

Furthermore, since it looks like `undefined` is being used as an error case, I think this is a prime case for throwing an exception.

With an exception, the function body simplifies to:

```js
const data = record.allMeta;
const firstVariant = data.visualVariants[0];
const nonvisualVariants = firstVariant.nonvisualVariants;

if (!nonVisualVariants[0]) throw new Error('No nonvisual variants');

return {
  data,
  firstVariant,
  nonvisualVariant,
};
```

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -100,14 +141,27 @@ export default class SaytDriverPlugin implements Plugin {
...
+   * Sends a request to the Search API with the given query and config object.
+   *
+   * @param query The search term to send.
+   * @param config Extra query-time configuration to customize the Search request.
+   * @returns A promise from the Search API that has been reformatted
+   * with the passed callback.
+   */
+  sendSearchApiRequest(query: string, config: QueryTimeAutocompleteConfig): Promise<ProductsResponseSection> {
+    const allConfig = { ...this.defaultSearchConfig, ...config };
+    return this.core.search.search({ query, ...allConfig }).then(this.searchCallback);
```
jmbtutor: I don't think `allConfig` enhances readability that much, so I think it can be inlined. As well, the default should be the first in the whole object so that any `query` parameters that might be given as default gets overridden.

```suggestion
    return this.core.search.search({ ...this.defaultSearchConfig, query, ...config }).then(this.searchCallback);
```

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -120,6 +174,55 @@ export default class SaytDriverPlugin implements Plugin {
...
+   * Extracts query and products from the given response.
+   *
+   * @param response An object containing the original query and product records.
+   * @returns An object containing the query and an array of valid simplified products.
+   */
+  searchCallback(response: Results): ProductsResponseSection {
+    const { query, records } = response;
+    const mappedRecords = records.map(record => {
+      const filter = this.parseRecord(record);
+      if (filter === undefined) return;
```
jmbtutor: See the comment for `parseRecord`.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -120,6 +174,55 @@ export default class SaytDriverPlugin implements Plugin {
...
+      const { data, firstVariant, nonvisualVariants } = filter;
+      return {
```
jmbtutor: We'll have to turn this into a product transformer later.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -120,6 +174,55 @@ export default class SaytDriverPlugin implements Plugin {
...
+        title: data.title,
+        price: nonvisualVariants[0].originalPrice,
+        imageSrc: firstVariant.productImage,
+        imageAlt: data.title,
+        productUrl: firstVariant.productImage,
+        // @TODO Handle variants
+      }
+    }).filter(record => record);
```
jmbtutor: Is `.filter(Boolean)` a common enough idiom for us to use?

packages/@sfx/sayt-driver-plugin/test/unit/common/sayt-driver-plugin.test.ts
```diff
@@ -227,59 +236,281 @@ describe('Sayt Driver Plugin', () => {
...
+
+      const result = driver.parseRecord(record);
+
+      expect(result).to.not.be.ok;
+    });
+  });
+
+  describe('searchCallback', () => {
+    let response;
+    beforeEach(() => {
```
jmbtutor: Newline above.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -120,6 +174,55 @@ export default class SaytDriverPlugin implements Plugin {
...
+  /**
+   * Extracts query and products from the given response.
+   *
+   * @param response An object containing the original query and product records.
+   * @returns An object containing the query and an array of valid simplified products.
+   */
+  searchCallback(response: Results): ProductsResponseSection {
+    const { query, records } = response;
+    const mappedRecords = records.map(record => {
+      const filter = this.parseRecord(record);
```
ChantalDesrochers: Is 'filter' an ideal name for this variable given it will sometimes return an object containing product data?

packages/@sfx/sayt-driver-plugin/test/unit/common/sayt-driver-plugin.test.ts
```diff
@@ -63,15 +70,16 @@ describe('Sayt Driver Plugin', () => {
   describe('ready()', () => {
     let registerListener;
 
-    it('should register an event listener for receiving sayt API requests', () => {
+    it('should register event listeners for receiving requests sayt and product data', () => {
```
ChantalDesrochers: should this say '... for sayt and product data requests'?

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
DanielMoniz-GBI: I like the exception idea. I do not really want to use an interface for this, however. Frankly, this code is so likely to change that adding an interface seems like make-work. It's a weird function to begin with. There will be a story to develop the transformation of products. At that time, we can update the relevant functions from this PR and create some meaningful interfaces.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -33,19 +35,41 @@ export default class SaytDriverPlugin implements Plugin {
...
   /**
    * Event name for sending Sayt autocomplete responses.
    */
-  saytResponseEvent: string = 'sfx::autocomplete_received_results';
+  autocompleteResponseEvent: string = 'sfx::autocomplete_received_results';
   /**
    * Event name to listen for Sayt autocomplete errors.
    */
-  saytErrorEvent: string = 'sfx::autocomplete_sayt_error';
+  autocompleteErrorEvent: string = 'sfx::autocomplete_sayt_error';
```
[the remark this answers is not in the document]
jmbtutor: Yes, the events ticket will be the one to normalize this. No action here.

packages/@sfx/sayt-driver-plugin/test/unit/common/sayt-driver-plugin.test.ts
```diff
@@ -227,59 +236,281 @@ describe('Sayt Driver Plugin', () => {
...
+          expectedGoodObject,
+        ],
+      };
+
+      const actualProduct = driver.searchCallback(input);
+
+      expect(actualProduct).to.deep.equal(expectedResponse);
+    });
+
+    it.skip('should return an object containing the query and products', () => {
```
jmbtutor: Remove `.skip`.

```suggestion
    it('should return an object containing the query and products', () => {
```

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -134,14 +237,26 @@ export default class SaytDriverPlugin implements Plugin {
   }
 }
 
+export interface RequestConfig {
+  query: string;
+  searchbox?: string;
+  config?: object;
```
jmbtutor: Since `config` is being overridden in the subtypes and you're not expecting a `RequestConfig` anywhere else, you don't need to include it here. Had there been a need for `config` in this interface, then the better approach would be to make `RequestConfig` generic whose type parameter is the type of `config`.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
jmbtutor: In that case, just `any` is sufficient.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
JoeyDeol: Maybe name the variable something like 'validRecord'?

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
ssbelle: @jmbtutor @DanielMoniz-GBI I see the same defaultSearchConfig object defined in both the sayt driver plugin and the search driver plugin.

ssbelle: @jmbtutor what is your suggested direction on this?

packages/@sfx/search-driver-plugin/src/search-driver-plugin.ts
```diff
@@ -71,7 +78,7 @@ export default class SearchDriverPlugin implements Plugin {
    * @param event the event whose payload is the search term.
    */
   fetchSearchData(event: CustomEvent<SearchRequestPayload>): void {
-    const searchTerm = event.detail;
+    const searchTerm = event.detail.value;
```
geoffcoutts: Shouldn't this also extract any config object from the view layer and send it to `sendSearchApiRequest()` like the `sayt-driver` does? Or I guess is that out of scope for this ticket?  Is there a ticket planned for this? Also, given that the functions will be mostly the same, should they be stored somewhere shared?

packages/@sfx/search-driver-plugin/src/search-driver-plugin.ts
```diff
@@ -86,13 +93,17 @@ export default class SearchDriverPlugin implements Plugin {
    *
    * @param query the query to send.
    */
-  sendSearchApiRequest(query: BridgeQuery): Promise<Results> {
-    return this.core.search.search(query);
+  sendSearchApiRequest(query: string): Promise<Results> {
+    const fullQuery = { ...this.defaultSearchConfig, query };
+    return this.core.search.search(fullQuery);
```
geoffcoutts: Going off the above ticket, this should prolly have some sort of callback to transform the results like the `sayt-driver` has, but again, prolly out of scope, correct?

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -33,19 +35,41 @@ export default class SaytDriverPlugin implements Plugin {
...
+   */
+  defaultSearchConfig: object = {
+    fields: ['*'],
+  }
 
   constructor() {
-    this.fetchSaytData = this.fetchSaytData.bind(this);
+    this.fetchAutocompleteTerms = this.fetchAutocompleteTerms.bind(this);
+    this.fetchProductData = this.fetchProductData.bind(this);
+    this.sendAutocompleteApiRequest = this.sendAutocompleteApiRequest.bind(this);
```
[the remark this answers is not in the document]
DanielMoniz-GBI: I'm of the mindset that a function needs to take care of its own interests. If a) this function is not usable without context, and b) we have no intention of purposely calling it with different contexts, we should bind it. It saves us in the future from bugs. I'm sure there are those that disagree with me.

Realistically, everything in a class should be automatically bound, but JavaScript wasn't written that way. Because we have to write it out method bindings manually, we are forced to choose which ones we bind. I'd prefer to choose not the minimal set for our current functionality, but a somewhat less minimal set that minimizes problems in the future.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
DanielMoniz-GBI: I'd assumed this would have to be more nuanced, but can you elaborate on what you mean?

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
DanielMoniz-GBI: Clever. I think the meaning would be pretty clear, even if people haven't seen it done before.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
DanielMoniz-GBI: I'm interested in your reasoning as to why `config?` shouldn't be in this interface. The listed attributes are indeed what we are expecting (whether mandatory or optional) in configuration for a request. Whether or not they are extended or used generically seems somewhat irrelevant.

The other argument I can see is that this interface exists entirely to remove duplication from the other two interfaces. From that point of view, it doesn't belong here. That being said, once it's created, it need not be the 'intent' of the interface. We can take the (despicable) post-modern point of view.

packages/@sfx/search-driver-plugin/src/search-driver-plugin.ts
DanielMoniz-GBI: This will be added in an upcoming ticket.

packages/@sfx/search-driver-plugin/src/search-driver-plugin.ts
DanielMoniz-GBI: Good question. We aren't really using search results right now in general, with the exception of the `SaytDriver` calling the `Search` plugin directly for use within the Sayt web component. Let's leave this out of scope until we have a good reason to change it.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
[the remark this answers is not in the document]
geoffcoutts: Yup, I'm fine if we want to bind everything as I don't think it's that costly. Although, I would think that if "b)" were true, then there's less reason to bind it.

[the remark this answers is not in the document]
DanielMoniz-GBI: What I meant by (b) was that there could be cases where you want a method to act differently depending on where it's called. The method's context would change, which you might want in that particular edge case. This almost never happens.

Keep in mind that simply running the following will cause the context to change:
```js
const fetchTerms = driver.fetchAutocompleteTerms;
fetchTerms();
```
This means that, whether or not we expect a method to be called in different places, it is very easy to accidentally supply a different context, such as in a test. That's why I think it's important to do.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
jmbtutor: Given the context, I think having `config?` here is correct (ignore my previous judgement), but it is more correctly typed with a generic instead of `object`.

```typescript
export interface RequestConfig<T> {
  query: string;
  searchbox?: string;
  config?: T;
}
```

The other interfaces can extend it like so:

```typescript
export interafce AutocompleteRequestConfig extends RequestConfig<QueryTimeAutocompleteConfig> {}
```

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
jmbtutor: By this, I meant that this transformation is specific to our sample dataset and that we'll have to make it more generic later.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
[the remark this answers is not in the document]
jmbtutor: My two problems with binding everything:

1. It doesn't allow for the context change at all, because a bound function cannot be rebound, and the default expectation to anyone who knows how `this` works is that a function is not bound.
2. It defeats the purpose of putting functions on the prototype because each copy of the component now gets its own set of functions. This can be costly, especially for components that have many instances.

Ideally, we would bind functions in the smallest necessary scope (i.e. at the place where it's passed down or assigned), but binding the function in the class constructor and putting it as an own property greatly simplifies testing and unregistering.

In my opinion, in the context of our components, we should only be binding functions that will be used as event listeners.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
jmbtutor: Is this not a `Partial<Request>`, `Request` coming from `groupby-api` (or in this case, `@sfx/search`? I haven't looked at the interface, though.

DanielMoniz-GBI: @jmbtutor I can't tell if you're asking me to change anything here.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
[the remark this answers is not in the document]
DanielMoniz-GBI: @jmbtutor
After thinking about this, given that the Sayt driver exposes no methods, we only have to worry about binding the event listeners. More generally, though, I disagree with your reasons (1) and (2).

For (1), if you can think of a reason why someone might want to rebind the function to another context, I'd be open to changing it. I've almost never seen this done on purpose in the wild. One exception is binding `this` to functions that are in the current scope (but automatically define their own context). Another exception is binding event listeners.

(2) It's not surprising to me that an instance receives its own set of functions. If you feel that 'costly' here is significant, I am open to changing it - I don't need you spending the time finding benchmarks. Given that it's pretty easy to accidentally remove the scope of a function you want to use in JavaScript, the simplest thing to do is bind the functions you're worried about. Additionally, we do not expect many instances of any of our plugins or drivers.

For the record, I never suggested binding everything. I suggested binding any function that uses `this`, and that's my default stance for code safety and predictability. I've run into issues around context binding enough times that I write programs around that concern defensively.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
jmbtutor: @DanielMoniz-GBI Typing something as `object` is generally incorrect, so if we have an interface for it readily available (which I suspect is `Partial<Request>`; the search plugin should be re-exporing `Request`), I'd use that instead.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -120,6 +170,57 @@ export default class SaytDriverPlugin implements Plugin {
...
+
+  /**
+   * Parses a given product record for valid data. Throws an error if data is invalid.
+   *
+   * @param record An object containing the product record data.
+   * @returns An object containing relevant product data.
+   */
+  parseRecord(record: Record): any {
+    const data = record.allMeta;
+    const firstVariant = data.visualVariants[0];
```
geoffcoutts: What is the standard convention that is put on record structures now? The apparel demo uses the naming convention you used here, Signet, BJs, and AB use `visualVariant` and `nonvisualVariant`, CVS uses `variants` and `subVariant`, and Zoro uses just `variants` to name a few. 

Is the assumption that a client would be able to override this and/or a product transformer that would be called at some point in the search response flow in order to fit their own data structure? 

I know this has been mentioned as needing to discuss with ETL and this doesn't need to be addressed right now, but flagging it here to raise attention to it.

DanielMoniz-GBI: @geoffcoutts Clients will need to be able to pass some sort of transformation that we can act on to provide the relevant product data. Or, the data needs to be in a standard format that we can use across all clients. The latter seems unlikely, but it's worth pushing for some sort of basic standardization.

The current solution is hardcoded for the `apparel` data. We'll need a story - possibly even more than one - to produce a solution for this. We can also refer to how Storefront dealt with this issue.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -120,6 +170,57 @@ export default class SaytDriverPlugin implements Plugin {
...
+      return {
+        title: data.title,
+        price: nonvisualVariants[0].originalPrice,
+        imageSrc: firstVariant.productImage,
+        imageAlt: data.title,
+        productUrl: firstVariant.productImage,
+        // @TODO Handle variants
+      }
+    }).filter(Boolean);
+    return {
```
jmbtutor: I guess it's fine for this case, but I want to be careful about doing too much processing and stripping down the data too much without giving access to the full response.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -134,14 +235,22 @@ export default class SaytDriverPlugin implements Plugin {
...
+  config?: T;
 }
+
+/**
+ * The type of the sayt autocomplete request event payload.
+ */
+export interface AutocompleteRequestConfig extends RequestConfig<QueryTimeAutocompleteConfig> {}
+
+/**
+ * The type of the sayt autocomplete request event payload.
```
jmbtutor: This description needs to change.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -157,3 +266,8 @@ export interface AutocompleteResponseSection {
 export interface SearchTermItem {
   label: string;
 }
+
+export interface ProductsResponseSection {
```
jmbtutor: Some basic documentation will help the events task. It doesn't have to be too refined because it will likely change with the events task.

packages/@sfx/sayt-driver-plugin/test/unit/common/sayt-driver-plugin.test.ts
```diff
@@ -227,59 +236,280 @@ describe('Sayt Driver Plugin', () => {
...
+      driver.sendSearchApiRequest(query, config);
+
+      expect(searchStub).to.be.calledWith({
+        query,
+        fields: ['*'],
+        ...config
+      });
+    });
+
+    it('should return the result of the Search API callback', () => {
```
jmbtutor:
```suggestion
    it('should resolve to the result of the Search API callback', () => {
```

packages/@sfx/sayt-driver-plugin/test/unit/common/sayt-driver-plugin.test.ts
```diff
@@ -227,59 +236,280 @@ describe('Sayt Driver Plugin', () => {
...
+
+      driver.fetchAutocompleteTerms(saytDataPayload);
 
-      driver.fetchSaytData(saytDataPayload);
+      return expect(Promise.resolve(dispatchEvent))
+        .to.be.eventually.calledOnceWith(driver.autocompleteErrorEvent, error);
+    });
+  });
+
+  describe('fetchProductData', () => {
```
jmbtutor: For consistency, add `()` to the end of the property name. This applies to the rest of the file as well.

```suggestion
  describe('fetchProductData()', () => {
```

packages/@sfx/sayt-driver-plugin/test/unit/common/sayt-driver-plugin.test.ts
```diff
@@ -227,59 +236,280 @@ describe('Sayt Driver Plugin', () => {
...
+          {
+            title: 'some-title',
+            price: 3.99,
+            imageSrc: 'some-link',
+            imageAlt: 'some-title',
+            productUrl: 'some-link',
+          },
+        ],
+      };
+
```
jmbtutor: Remove newline to group the setup together.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
jmbtutor: I would add a newline above for readability.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -120,6 +170,57 @@ export default class SaytDriverPlugin implements Plugin {
...
   }
 
+  /**
+   * Extracts query and products from the given response.
+   *
+   * @param response An object containing the original query and product records.
+   * @returns An object containing the query and an array of valid simplified products.
+   */
+  searchCallback(response: Results): ProductsResponseSection {
+    let filter;
```
jmbtutor: This variable is only used inside the `map` callback, so we can define it there instead to minimize its lifetime.

packages/@sfx/search-driver-plugin/test/unit/common/search-driver-plugin.test.ts
```diff
@@ -72,7 +72,7 @@ describe('SearchDriverPlugin', () => {
         [eventsPluginName]: { dispatchEvent: () => {} },
       };
 
-      searchDriverPlugin.fetchSearchData({ detail: searchTerm } as any);
+      searchDriverPlugin.fetchSearchData({detail: { value: searchTerm } } as any);
```
jmbtutor:
```suggestion
      searchDriverPlugin.fetchSearchData({ detail: { value: searchTerm } } as any);
```

packages/@sfx/search-driver-plugin/src/search-driver-plugin.ts
```diff
@@ -28,6 +28,13 @@ export default class SearchDriverPlugin implements Plugin {
    */
   eventsPluginName: string = 'dom_events';
 
+  /**
+   * Default configuration for all searches.
+   */
+  defaultSearchConfig: object = {
```
jmbtutor: I think this is also a `Partial<Request>`.

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
DanielMoniz-GBI: Yeah, I'm unsure of whether or not we want to filter anything at all. In the meantime, this function works for the apparel data and will be updated to be more complete (even before adding a proper client-friendly transformer).

packages/@sfx/sayt-driver-plugin/src/sayt-driver-plugin.ts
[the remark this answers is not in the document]
jmbtutor: Agree to disagree on the finer points, but I think we agree from a practical standpoint, so I'm okay to resolve this.

PR #27, opened 2019-09-13
SFX-198: Implement plugin unregistration

jmbtutor:
A new method `Core.unregister` has been added to unregister a subset of plugins. It takes in a list of plugins to unregister and unregisters them. A new dependency graph structure has been added to keep track of dependency relationships, and `Core.register` has been modified to populate that graph. The graph is represented as an adjacency list, where the keys are plugin names and the values are arrays of dependers. A number of dependency graph-related helper functions have also been created.

### DIFF
--- a/packages/@sfx/core/src/core.ts
+++ b/packages/@sfx/core/src/core.ts
@@ -9,6 +9,7 @@ import {
 import {
   DependencyGraph,
   createDependencyGraph,
+  mergeDependencyGraphs,
 } from './utils/dependencies';
 
 /**
@@ -74,6 +75,7 @@ export default class Core {
     }
 
     registerPlugins(plugins, this.registry, this.directory);
+    this.dependencyGraph = mergeDependencyGraphs(this.dependencyGraph, createDependencyGraph(plugins));
 
     initPlugins(plugins);
     readyPlugins(plugins);
--- a/packages/@sfx/core/src/core.ts
+++ b/packages/@sfx/core/src/core.ts
@@ -6,6 +6,10 @@ import {
   registerPlugins,
   unregisterPlugins,
 } from './utils/core';
+import {
+  DependencyGraph,
+  createDependencyGraph,
+} from './utils/dependencies';
 
 /**
  * The core of the SF-X plugin system. This entity is responsible for
@@ -30,6 +34,13 @@ export default class Core {
    * Plugins do not have access to this directory.
    */
   directory: PluginDirectory = Object.create(null);
+
+  /**
+   * An object with plugin names mapped to the names of the plugins that
+   * depend on it. This dependency graph is represented as an adjacency
+   * list.
+   */
+  dependencyGraph: DependencyGraph = createDependencyGraph();
 
   /**
    * Register one or more plugins with Core.
--- a/packages/@sfx/core/src/core.ts
+++ b/packages/@sfx/core/src/core.ts
@@ -81,6 +81,10 @@ export default class Core {
     readyPlugins(plugins);
   }
 
+  unregister(plugins: string[]) {
+    unregisterPlugins(plugins, this.registry, this.directory);
+  }
+
   /**
    * Unregisters all plugins. The plugin registry, directory and
    * dependency graph are all cleared. The optional `unregister`
--- a/packages/@sfx/core/src/core.ts
+++ b/packages/@sfx/core/src/core.ts
@@ -10,6 +10,7 @@ import {
   DependencyGraph,
   createDependencyGraph,
   mergeDependencyGraphs,
+  removeFromDependencyGraph,
 } from './utils/dependencies';
 
 /**
@@ -83,6 +84,7 @@ export default class Core {
 
   unregister(plugins: string[]) {
     unregisterPlugins(plugins, this.registry, this.directory);
+    this.dependencyGraph = removeFromDependencyGraph(this.dependencyGraph, plugins);
   }
 
   /**
--- a/packages/@sfx/core/src/core.ts
+++ b/packages/@sfx/core/src/core.ts
@@ -85,7 +85,7 @@ export default class Core {
   /**
    * Unregisters the given plugins. The plugin registry, directory and
    * dependency graph are all updated. The optional `unregister`
-   * function of each plugin is called with the plugin is unregistered.
+   * function of each plugin is called when the plugin is unregistered.
    * The order in which the plugins are unregistered is unspecified.
    *
    * @param plugins The names of the plugins to unregister.
--- a/packages/@sfx/core/src/core.ts
+++ b/packages/@sfx/core/src/core.ts
@@ -83,8 +83,9 @@ export default class Core {
   }
 
   unregister(plugins: string[]) {
+    const updatedGraph = removeFromDependencyGraph(this.dependencyGraph, plugins);
     unregisterPlugins(plugins, this.registry, this.directory);
-    this.dependencyGraph = removeFromDependencyGraph(this.dependencyGraph, plugins);
+    this.dependencyGraph = updatedGraph;
   }
 
   /**
--- a/packages/@sfx/core/src/core.ts
+++ b/packages/@sfx/core/src/core.ts
@@ -82,6 +82,15 @@ export default class Core {
     readyPlugins(plugins);
   }
 
+  /**
+   * Unregisters the given plugins. The plugin registry, directory and
+   * dependency graph are all updated. The optional `unregister`
+   * function of each plugin is called with the plugin is unregistered.
+   * The order in which the plugins are unregistered is unspecified.
+   *
+   * @param plugins The names of the plugins to unregister.
+   * @throws If unregistering a plugin will break a dependency.
+   */
   unregister(plugins: string[]) {
     const updatedGraph = removeFromDependencyGraph(this.dependencyGraph, plugins);
     unregisterPlugins(plugins, this.registry, this.directory);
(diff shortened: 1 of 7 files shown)


packages/@sfx/core/src/utils/dependencies.ts
```diff
@@ -0,0 +1,93 @@
...
+}
+
+/**
+ * Creates an empty dependency graph.
+ *
+ * @returns An empty dependency graph.
+ * @hidden
+ */
+function createEmptyGraph(): DependencyGraph {
+  return Object.create(null);
```
DanielMoniz-GBI: Any reason why this shouldn't be an empty array? I would consider that the 'trivial' dependency graph - there are no plugins or dependencies.

packages/@sfx/core/src/core.ts
```diff
@@ -31,6 +37,13 @@ export default class Core {
    */
   directory: PluginDirectory = Object.create(null);
 
+  /**
+   * An object with plugin names mapped to the names of the plugins that
+   * depend on it. This dependency graph is represented as an adjacency
+   * list.
+   */
+  dependencyGraph: DependencyGraph = createDependencyGraph();
```
DanielMoniz-GBI: High-level question: do we need a dependency graph if we already have a list of plugins, each plugin having its own list of plugins on which it depends? An alternative would be creating utilities for navigating our existing plugins and their dependencies without explicitly building a graph.

jmbtutor: For posterity: after some discussion, we concluded that there is a way to do it without building the graph. But also after discussion, we decided to leave this implementation for time's sake.

The alternate solution is as follows:

1. Create a hash (`Set`, `Object`, `Map`; something with O(1) lookup) from the plugins to remove.
2. For each plugin in the registry, excluding the ones to be removed, check that it has none of the plugins to remove listed in its dependencies. If it does, add the plugin's name to a structure that associates the name with the dependency.
3. If the dependencies are clean, the check passes. If not, throw with an appropriate error with a message that includes the dependencies and its dependers.

Worst case, this solution is O(n^2), but on average, it will probably be O(n).

packages/@sfx/core/src/utils/dependencies.ts
DanielMoniz-GBI: Yeah, I realized it on second review today. Thanks!

packages/@sfx/core/test/unit/common/core.test.ts
```diff
@@ -113,5 +204,14 @@ describe('Core', () => {
...
       );
     });
+
+    it('should clear dependency graph', () => {
+      const unregisterPlugins = stub(CoreUtils, 'unregisterPlugins');
+      core.dependencyGraph = { a: ['a'] };
+
+      core.unregisterAll();
+
+      expect(core.dependencyGraph).to.be.empty;
```
DanielMoniz-GBI: I think I like this. No need to stub an internal detail like a call to `createDependencyGraph()`. Unless, that is, we expect an 'empty' dependency graph to be created with more information than an empty object.

packages/@sfx/core/test/unit/common/utils/core.test.ts
```diff
@@ -476,5 +476,33 @@ describe('CoreUtils', () => {
 
       expect(() => unregisterPlugins(names, registry, directory)).to.not.throw();
     });
+
+    it('should ignore plugins not in the directory', () => {
```
DanielMoniz-GBI: This is a bit of an odd test. I guess you're asserting that the function acts as normal if non-existent plugins are passed. Is there something simpler we can assert? Eg. that it doesn't throw an error? This is fine, but it looks like it's pretty much the same as the test above. Or maybe we can just wrap them into one test.

packages/@sfx/core/test/unit/common/utils/dependencies.test.ts
```diff
@@ -0,0 +1,258 @@
...
+
+      const dependencies = createDependencyGraph(plugins);
+
+      expect(dependencies).to.deep.equal({
+        a: ['b'],
+        b: ['a'],
+        c: ['c'],
+      });
+    });
+  });
```
DanielMoniz-GBI: Nice test suite above.

packages/@sfx/core/test/unit/common/utils/dependencies.test.ts
```diff
@@ -0,0 +1,258 @@
...
+
+  describe('mergeDependencyGraphs()', () => {
+    it('should return an object without inherited properties', () => {
```
DanielMoniz-GBI: Can you elaborate on the specific reason(s) why we're avoiding inherited properties?

packages/@sfx/core/test/unit/common/utils/dependencies.test.ts
```diff
@@ -0,0 +1,258 @@
...
+      const dependersOfPluginA = [];
+      const graph = {
+        a: dependersOfPluginA,
+        b: ['a', 'c'],
+        c: [],
+      };
+
+      const newGraph = removeFromDependencyGraph(graph, ['a']);
+
+      expect(newGraph).not.to.equal(graph);
```
DanielMoniz-GBI: Is this enough? Should we assert that the new graph does not have a key of `a`?

packages/@sfx/core/src/utils/dependencies.ts
```diff
@@ -0,0 +1,93 @@
...
+ * corresponding arrays are concatenated.
+ *
+ * @param graphs The dependency graphs to merge.
+ * @returns The merged dependency graph.
+ */
+export function mergeDependencyGraphs(...graphs: DependencyGraph[]): DependencyGraph {
+  return graphs.reduce((merged, graph) => {
+    Object.keys(graph).forEach((key) => {
+      if (!merged[key]) merged[key] = [];
+      merged[key].push(...graph[key]);
```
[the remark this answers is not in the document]
jmbtutor: That was intentional, but we can discuss if it's correct. That said, since plugins must have unique names, I don't think anything will break if I deduplicate.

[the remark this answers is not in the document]
DanielMoniz-GBI: As it stands, I'm guessing an error message would be confusing if the dependency is in the same array more than once. Not a huge deal, but it's a bit odd.

packages/@sfx/core/test/unit/common/utils/core.test.ts
DanielMoniz-GBI: It's fine and is not worth spending too much time.

packages/@sfx/core/test/unit/common/utils/dependencies.test.ts
DanielMoniz-GBI: It is, but I think it would make this test much clearer. Effectively:
```
expect oldGraph to have key 'a'
expect newGraph to not have key 'a'
```
This tells a nice little story.

packages/@sfx/core/test/unit/common/core.test.ts
```diff
@@ -113,5 +204,14 @@ describe('Core', () => {
         sinon.match.same(directory)
       );
     });
+
+    it('should clear dependency graph', () => {
+      const unregisterPlugins = stub(CoreUtils, 'unregisterPlugins');
```
ChantalDesrochers: Does this need to be stored in a variable if it's not being used?

packages/@sfx/core/src/utils/dependencies.ts
```diff
@@ -0,0 +1,97 @@
...
+ * Builds a dependency graph from the given plugins.
+ *
+ * @param plugins The plugins from which to build the graph.
+ * @returns The completed dependency graph.
+ */
+export function createDependencyGraph(plugins: Plugin[] = []): DependencyGraph {
+  const dependencyGraph = createEmptyGraph();
+
+  plugins.forEach(({ metadata: { name, depends } }) => {
+    if (!dependencyGraph[name]) dependencyGraph[name] = [];
```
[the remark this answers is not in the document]
jmbtutor: If we use the dependency graph for any analysis later on, it's a lot simpler down the line for the caller to be able to assume that a plugin exists rather than having to check that the plugin exists in the graph before checking its dependers. It's also a simpler invariant to say that the graph contains all plugins rather than all plugins that have dependers.

packages/@sfx/core/src/core.ts
```diff
@@ -63,18 +76,35 @@ export default class Core {
...
-   * plugins are unregistered is unspecified.
+   * Unregisters the given plugins. The plugin registry, directory and
+   * dependency graph are all updated. The optional `unregister`
+   * function of each plugin is called with the plugin is unregistered.
+   * The order in which the plugins are unregistered is unspecified.
+   *
+   * @param plugins The names of the plugins to unregister.
+   * @throws If unregistering a plugin will break a dependency.
+   */
+  unregister(plugins: string[]) {
```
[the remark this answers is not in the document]
jmbtutor: I wanted to keep it consistent with the rest of the file and have the linting ticket take care of it, but I can add this in.

packages/@sfx/core/test/unit/common/core.test.ts
jmbtutor: It does not. I'll remove it.

packages/@sfx/core/src/core.ts
```diff
@@ -63,18 +76,35 @@ export default class Core {
...
   }
 
   /**
-   * Unregisters all plugins. The plugin registry and directory are both
-   * cleared. The optional `unregister` function of each plugin is
-   * called when the plugin is unregistered. The order in which the
-   * plugins are unregistered is unspecified.
+   * Unregisters the given plugins. The plugin registry, directory and
+   * dependency graph are all updated. The optional `unregister`
+   * function of each plugin is called with the plugin is unregistered.
```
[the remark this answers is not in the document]
jmbtutor: Yes, that should say "when". The sentence refers to the `unregister` function of each plugin, as stated in the first half of the sentence.

PR #28, opened 2019-09-16
SFX-354: Update search driver for searchbox IDs

geoffcoutts:
The search driver is receiving an optional searchbox ID in search requests. This value should be returned in the product response events that the search driver sends.

### DIFF
--- a/packages/@sfx/search-driver-plugin/src/search-driver-plugin.ts
+++ b/packages/@sfx/search-driver-plugin/src/search-driver-plugin.ts
@@ -83,8 +83,8 @@ export default class SearchDriverPlugin implements Plugin {
       .then((results) => {
         this.core[this.eventsPluginName].dispatchEvent(SEARCH_RESPONSE_EVENT, { results, searchbox });
       })
-      .catch((e) => {
-        this.core[this.eventsPluginName].dispatchEvent(SEARCH_ERROR_EVENT, { e, searchbox });
+      .catch((error) => {
+        this.core[this.eventsPluginName].dispatchEvent(SEARCH_ERROR_EVENT, { error, searchbox });
       });
   }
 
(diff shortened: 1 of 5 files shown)


packages/@sfx/search-driver-plugin/src/search-driver-plugin.ts
```diff
@@ -78,10 +78,10 @@ export default class SearchDriverPlugin implements Plugin {
...
   fetchSearchData(event: CustomEvent<SearchRequestPayload>): void {
-    const searchTerm = event.detail.value;
+    const { value: searchTerm, searchbox } = event.detail;
     this.sendSearchApiRequest(searchTerm)
       .then((results) => {
-        this.core[this.eventsPluginName].dispatchEvent(SEARCH_RESPONSE_EVENT, results);
+        this.core[this.eventsPluginName].dispatchEvent(SEARCH_RESPONSE_EVENT, { results, searchbox });
       })
       .catch((e) => {
         this.core[this.eventsPluginName].dispatchEvent(SEARCH_ERROR_EVENT, e);
```
DanielMoniz-GBI: We haven't been using the error events yet, but given that an error would relate to the request from a specific searchbox, the error event should probably contain the `searchbox` as well. That could helps client place the error on the page, for example.

Note that the same thing applies to the error events in the `SaytDriver`.

geoffcoutts: I saw that we weren't sending it back in the `SaytDriver`, but it probably make sense to be added. The events repo will be including it as well so I can change this to 
```suggestion
        this.core[this.eventsPluginName].dispatchEvent(SEARCH_ERROR_EVENT, { e, searchbox });
```

Would you like me to do the same for the `SaytDriver`?

DanielMoniz-GBI: Yes please. Given that this story technically touches both - you just didn't need to touch the `SaytDriver` before - I think it makes sense to be part of this ticket.

PR #29, opened 2019-09-16
SFX 291 - Add linter

ChantalDesrochers:
- add linter to lint all files under /src and /test
- create linting scripts at the package and root levels
- create scripts to fix linting errors that can be automatically fixed
- incorporate lint checks into CircleCi
- NOTE: open to discussion in terms of linting rule choices (overrides, removals, etc.)
- NOTE2: disabled next line on most imports, because a rule requires imported packages (e.g. autocomplete into sayt) be listed as a dependency, however this is not currently possible
- NOTE3: in some instances, converted imports of events names to use paths rather than packages - this will be resolved with the events repo integration


.eslintrc.js
```diff
@@ -0,0 +1,62 @@
+module.exports = {
```
jmbtutor: I think this file is basically the same as what's in the View repo, so I think we can move this to the devtools repo instead.

packages/@sfx/core/test/unit/common/index.test.ts
```diff
@@ -14,14 +14,17 @@ describe('Entry point', () => {
   });
 
   it('should export the Plugin interface', () => {
+    // eslint-disable-next-line
```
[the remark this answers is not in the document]
ChantalDesrochers: @typescript-eslint/no-unused-vars (I added this specification to the disable comment)

test/.eslintrc.js
```diff
@@ -0,0 +1,37 @@
...
+    parser: "@typescript-eslint/parser",
+    parserOptions: {
+        ecmaVersion: 2018,
+        sourceType: "module",
+        tsconfigRootDir: __dirname,
+        project: [
+            path.resolve(__dirname, "../tsconfig.json"),
+          ]
+        },
+    rules: {
```
jmbtutor: Can we put these rules into the devtools repo and have this (and the corresponding file in the View repo) extend it?

PR #34, opened 2019-10-25
SFX-158: Add Cache and Cache Driver plugins

jmbtutor:


packages/@sfx/cache-driver-plugin/src/index.ts
```diff
@@ -0,0 +1,2 @@
+// eslint-disable-next-line import/prefer-default-export
```
DanielMoniz-GBI: I needed to add this `eslint-disable` line because there is a single export below and, by our convention, it is not a default export.

packages/@sfx/cache-driver-plugin/test/unit/common/cache-driver-plugin.test.ts
```diff
@@ -0,0 +1,79 @@
...
+      const registry: any = { a: 1, b: 2 };
+
+      cacheDriverPlugin.register(registry);
+
+      expect(cacheDriverPlugin.core).to.equal(registry);
+    });
+  });
+
+  describe('ready()', () => {
+    it(`should listen for ${CACHE_REQUEST}`, () => {
```
DanielMoniz-GBI: I'm not sure I'm sold on using dynamic test descriptions, as this adds an extra level of indirection and makes descriptions harder to search. The developer does not even need to know the actual name of the event.

packages/@groupby/elements-sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -126,6 +127,7 @@ export default class SaytDriverPlugin<P = Record> implements Plugin {
       .then((results) => {
         const payload: AutocompleteResponsePayload = { results, group };
         this.core[this.eventsPluginName].dispatchEvent(AUTOCOMPLETE_RESPONSE, payload);
+        if (this.core.cache) this.core.cache.set(`${AUTOCOMPLETE_RESPONSE}::${group}`, payload);
```
geoffcoutts: Update the plugin's changelog to reflect this?

packages/@groupby/elements-cache-plugin/src/cache-plugin.ts
```diff
@@ -0,0 +1,54 @@
...
+    return {
+      name: 'cache',
+      depends: [],
+    };
+  }
+
+  /**
+   * The data store in which cache entries are stored.
+   */
+  store: Map<unknown, unknown>;
```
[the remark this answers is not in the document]
jmbtutor: We know what the types are for our specific use cases, but since this is a general plugin that other plugins can use, we don't know it in the general case. `unknown, unknown` is likely the most correct type here.

packages/@groupby/elements-cache-plugin/src/cache-plugin.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  /**
+   * Instantiates this plugin. By default, an empty `Map` is created as
+   * the data store.
+   *
+   * @param __namedParameters Options for plugin configuration.
```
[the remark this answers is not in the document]
jmbtutor: Yes.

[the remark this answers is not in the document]
jmbtutor: Yes.

packages/@groupby/elements-cache-driver-plugin/src/cache-driver-plugin.ts
```diff
@@ -0,0 +1,61 @@
...
+  ready(): void {
+    this.core.dom_events.registerListener(CACHE_REQUEST, this.handleRequest);
+  }
+
+  /**
+   * Dispatches the requested data from the cache.
+   *
+   * @param req The cache request details.
+   */
+  handleRequest(req: CustomEvent<CacheRequestPayload>): void {
```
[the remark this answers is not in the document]
jmbtutor: `CACHE_ERROR` was for the case where there is an error between this plugin requesting data from the cache and the cache returning the requested data. I guess we can't guarantee that the cache is what it claims to be, so perhaps it would be good to wrap it in a `try..catch` in case there are any errors. After a discussion with @DanielMoniz-GBI, whether or not we decide to do it, we will push it off to a later date and this PR can go through without it.

packages/@groupby/elements-cache-driver-plugin/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+This plugin listens for and dispatches a number of events. These events are defined in the [`@groupby/elements-events`][elements-events] package.
+
+### Received
+
+* `CACHE_REQUEST`: When received, a request to the cache is made and a response is dispatched with the event name provided in the payload of `CACHE_REQUEST`.
+
+### Dispatched
+
+* A cache response event: Dispatched when a `CACHE_REQUEST` is fulfilled. The event is dispatched under the name provided with the corresponding `CACHE_REQUEST` payload.
```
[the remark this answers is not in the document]
jmbtutor: > Is it also worth including an example of what one of these event names could look like?

I can add a recommendation about what it should look like for the `CACHE_REQUEST` event, but we have no idea what it would look like for `CACHE_RESPONSE`. I think more extensive documentation might be better suited elsewhere.

packages/@groupby/elements-cache-driver-plugin/src/cache-driver-plugin.ts
```diff
@@ -0,0 +1,63 @@
...
+   */
+  register(plugins: PluginRegistry): void {
+    this.core = plugins;
+  }
+
+  /**
+   * Registers a number of event listeners.
+   * The following events are listened for:
+   *
+   * - [[CACHE_REQUEST]]
```
DanielMoniz-GBI: This feels unnecessary when we have only one item, but it's not a big deal.

packages/@groupby/elements-cache-driver-plugin/src/cache-driver-plugin.ts
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  /**
+   * Dispatches the requested data from the cache.
+   * The event name provided by the request is used as the name of the
+   * return event.
+   *
+   * @param req The cache request details.
+   */
+  handleRequest(req: CustomEvent<CacheRequestPayload>): void {
```
DanielMoniz-GBI: I would really prefer the variable to be called `request`. It's not really that long of a name.

packages/@groupby/elements-cache-driver-plugin/test/unit/common/cache-driver-plugin.test.ts
```diff
@@ -0,0 +1,87 @@
...
+    it('should dispatch the requested data from the cache', () => {
+      const req = { detail: { name, returnEvent, group } };
+      cache.set('key::grp', data);
+
+      cacheDriverPlugin.handleRequest(req);
+
+      expect(dispatchEvent).to.be.calledWith(returnEvent, { name, data, group });
+    });
+
+    it('should expand an undefined group to the empty string', () => {
```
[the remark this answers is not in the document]
DanielMoniz-GBI: Yes. Specifically, asserting that `'key::'` is the used key in the cache implies that the `undefined` value became an empty string.

packages/@groupby/elements-cache-driver-plugin/test/unit/common/cache-driver-plugin.test.ts
```diff
@@ -0,0 +1,87 @@
...
+      const req = { detail: { name, returnEvent } };
+      cache.set('key::', data);
+
+      cacheDriverPlugin.handleRequest(req);
+
+      expect(dispatchEvent).to.be.calledWith(returnEvent, { name, data, group: undefined });
+    });
+
+    it('should dispatch an undefined payload if no data is cached', () => {
```
DanielMoniz-GBI: I'm starting to question whether this is a good idea. In what case would we want to initialize a component with an empty/undefined payload? Taking this action makes me think that we're opening ourselves up to future bugs where the component becomes empty when we didn't want it to.

I don't believe it's causing any bugs at the moment, and this is a valid choice, but I am questioning it.

packages/@groupby/elements-cache-plugin/src/cache-plugin.ts
```diff
@@ -0,0 +1,54 @@
...
+    return this.store;
+  }
+}
+
+/**
+ * The type of the [[CachePlugin]] options object.
+ */
+export interface CachePluginOptions {
+  /** The store with which to initialize the plugin. */
+  store: Map<unknown, unknown>;
```
DanielMoniz-GBI: Given that we use `Map<unknown, unknown>` three times, should we have a specific type for this and reuse it? Say, `CacheMap`?

packages/@groupby/elements-cache-driver-plugin/test/unit/common/cache-driver-plugin.test.ts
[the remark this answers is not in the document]
geoffcoutts: What I thought, but the test description seemed a little ambiguous to me at the time.

packages/@groupby/elements-cache-driver-plugin/test/unit/common/cache-driver-plugin.test.ts
jmbtutor: The trick is that we're not initializing the component with an empty/undefined payload. It's the responsibility of the receiver to handle an empty cache response and other unexpected data. In our case, if we get an empty response in the autocomplete component, we default to an empty array for results.

That said, I originally didn't have this as a test case. The behaviour I specified in the initial set of test cases was that the plugin forwarded what the cache provided. It just happens that the cache responds with `undefined` when the key doesn't exist.

packages/@groupby/elements-cache-driver-plugin/src/cache-driver-plugin.ts
jmbtutor: I'll leave it as-is for consistency and for ease of future modification.

packages/@groupby/elements-cache-plugin/src/cache-plugin.ts
```diff
@@ -0,0 +1,59 @@
...
+ */
+export interface CachePluginOptions {
+  /** The store with which to initialize the plugin. */
+  store: CacheStore;
+}
+
+/**
+ * The type of the cache store.
+ */
+export type CacheStore = Map<unknown, unknown>;
```
jmbtutor: We should consider turning off [`@typescript-eslint/no-empty-interface`](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-empty-interface.md), which prevented me from using an interface like so:

```typescript
export interface CacheStore extends Map<unknown, unknown> {}
```

PR #35, opened 2019-11-01
SFX-437: Rename project to 'Elements'

ChantalDesrochers:
Notes:
- currently linking to elements-events and elements-devtools SFX-437 PRs
- back and forth between what string to use for the module use options, with expose loader. I went with 'elementsCore' and 'elementsPlugins' because gbe was used so rarely on the logic layer side, it felt a bit foreign, but also very strongly considered 'gbePlugins'...
UPDATE: currently GbElementsCore & GbElementsPlugins


packages/@elements/search-driver-plugin/package.json
```diff
@@ -19,12 +19,12 @@
...
-    "url": "https://github.com/groupby/sfx-logic.git",
-    "directory": "packages/@sfx/search-driver-plugin"
+    "url": "https://github.com/groupby/elements-logic.git",
+    "directory": "packages/@elements/search-driver-plugin"
   },
   "author": "",
   "license": "MIT",
   "dependencies": {
-    "@sfx/events": "groupby/sfx-events#c4a4853a424992db9c170fea6990dc2a52c3061e"
+    "@elements/events": "groupby/elements-events#c4a4853a424992db9c170fea6990dc2a52c3061e"
```
ssbelle: question does this hash change at this point? Or do we expect it to change once we update the events repo?

packages/@elements/sayt-driver-plugin/package.json
```diff
@@ -19,13 +19,13 @@
...
-    "directory": "packages/@sfx/sayt-driver-plugin"
+    "url": "https://github.com/groupby/elements-logic.git",
+    "directory": "packages/@elements/sayt-driver-plugin"
   },
   "author": "",
   "license": "MIT",
   "sideEffects": false,
   "dependencies": {
-    "@sfx/events": "groupby/sfx-events#c4a4853a424992db9c170fea6990dc2a52c3061e"
+    "@elements/events": "groupby/elements-events#c4a4853a424992db9c170fea6990dc2a52c3061e"
```
ssbelle: same here, should this hash be different?

packages/@elements/search-driver-plugin/package.json
ChantalDesrochers: I used yarn link to use the PR that introduces the renaming for now. Once that is merged in, I will be updating the hash. (if I'm remembering the process correctly)

webpack.config.js
```diff
@@ -21,11 +21,11 @@ module.exports = {
   module: {
     rules: [{
       test: path.resolve(__dirname, 'presets', 'core.ts'),
-      use: { loader: 'expose-loader', options: 'sfxCore' }
+      use: { loader: 'expose-loader', options: 'GbElementsCore' }
```
DanielMoniz-GBI: If we're keeping the same convention as before, this new object should be called `gbElementsCore` instead of `If we're keeping the same convention as before, this new object should be called `GbElementsCore`. Is there any reason for the new capitalization?

(I noticed this because I'm updating the demo in the View Layer and needed to instantiate Core)

.circleci/config.yml
```diff
@@ -40,9 +40,9 @@ defaults:
       cp .circleci/prepare-commit-msg .git/hooks/prepare-commit-msg
       chmod +x .git/hooks/prepare-commit-msg
 
-  requires_build_core: &requires_build_core
+  requires_build_elements_core: &requires_build_elements_core
     requires:
-      - build_core
+      - build_elements_core
```
jmbtutor: Do we expect any packages that don't start with `elements-`? If not, we can drop `elements` from the job names and add the prefix where we use it instead. That should also significantly reduce the diff for this file.

.circleci/config.yml
```diff
@@ -95,7 +95,7 @@ jobs:
             package="${CIRCLE_JOB#build_}"
             package="${package//_/-}"
 
-            cd "packages/@sfx/${package}"
+            cd "packages/@groupby/${package}"
```
jmbtutor: If you follow the suggestion above and drop `elements` from all the job names, this and similar lines will change to the following:

```suggestion
            cd "packages/@groupby/elements-${package}"
```

webpack.config.js
jmbtutor: I never starting the names of objects used as "variable bags" with a lowercase letter. It's not really a value that you'd pass around and use by itself: you'd always access some property of it. We have `Math`, `JSON`, `Symbol`, and [others](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Not_a_constructor) as precedents in favour of capitalizing the first letter.

PR #38, opened 2019-11-13
ELE-453: Allow for caching of products

ssbelle:


packages/@groupby/elements-search-driver-plugin/test/unit/common/search-driver-plugin.test.ts
```diff
@@ -85,45 +85,56 @@ describe('SearchDriverPlugin', () => {
...
+    let config;
     let results;
     let group;
     let sendSearchApiRequest;
+    let query;
 
     beforeEach(() => {
-      results = { a: 'a' };
+      area = 'area';
+      collection = 'collection';
```
jmbtutor: Since these are strings (and therefore primitives), you can turn the variables into `const`s and initialize them outside of the `beforeEach`.

PR #44, opened 2019-12-05
ELE-249: Create first pass of tracker service plugin for Search

DanielMoniz-GBI:
* Add `GbTrackerPlugin` for automatic GroupBy beaconing integration.
* Update `SaytDriverPlugin` to dispatch a search beacon after a successful product search.
* Update `SearchDriverPlugin` to dispatch a search beacon after a successful search.


packages/@groupby/elements-gb-tracker-plugin/src/gb-tracker-plugin.ts
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * The options used to instantiate a [[GbTrackerPlugin]].
+ */
+export interface TrackerPluginOptions {
+  /** The ID of the client. */
+  customerId: string;
+  /** The area in which the tracked actions will occur. */
+  area?: string;
+  // @TODO Is this option needed? What does it do?
+  overridePixelUrl?: string;
```
DanielMoniz-GBI: I still need to figure out what this is used for. I can't find documentation on it.

packages/@groupby/elements-search-driver-plugin/src/search-driver-plugin.ts
```diff
@@ -137,6 +148,21 @@ export default class SearchDriverPlugin<P = Record> implements Plugin {
...
+   * @param originValue The search event's origin.
+   */
+  dispatchSearchTrackerEvent(results: Results, originValue: string): void {
+    const origin: SendableOrigin = { [originValue]: true };
+    const beaconSearchPayload: BeaconSearchPayload = {
+      results,
+      origin,
+    };
+    this.core[this.eventsPluginName].dispatchEvent(BEACON_SEARCH, beaconSearchPayload);
+  }
```
DanielMoniz-GBI: It's worth noting that this is completely duplicated between the `SearchDriverPlugin` and the `SaytDriverPlugin`. I believe there is other functionality as well. We should think about how our plugins can share functionality.

packages/@groupby/elements-sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -136,7 +140,12 @@ export default class SaytDriverPlugin<P = Record> implements Plugin {
    * @param event Event that contains the Search API request payload.
    */
   fetchProductData(event: CustomEvent<SaytProductsRequestPayload>): void {
-    const { query, group, config } = event.detail;
+    const {
+      query,
+      group,
+      config,
+      origin,
+    } = event.detail;
```
DanielMoniz-GBI: This could easily be one line, but I believe the linter complains if you import >= 4 modules on one line. Personally, I think that should just be a line length issue.

packages/@groupby/elements-search-driver-plugin/test/unit/common/search-driver-plugin.test.ts
```diff
@@ -165,11 +175,21 @@ describe('SearchDriverPlugin', () => {
...
 
-      searchDriverPlugin.fetchSearchData({ detail: 'search' } as any);
+    it('should dispatch a search tracker event with origin when the search succeeds', (done) => {
+      const origin = 'some-origin';
+      response.originalResponse = { id: 'search-id' };
+      sendSearchApiRequest.resolves(response);
+      dispatchSearchBeacon.callsFake(() => {
+        expect(dispatchSearchBeacon).to.be.calledWith(response.originalResponse, origin);
+        done();
+      });
```
DanielMoniz-GBI: This is notably a different style of test from that of the `SaytDriverPlugin`. These are more finicky, as additionally calls to the relevant fake end up triggering multiple `done()` calls. Not ideal. That being said, I don't think this is the time to refactor the rest of the file. Feel free to add any thoughts below.

packages/@groupby/elements-gb-tracker-plugin/README.md
```diff
@@ -0,0 +1,44 @@
...
+    collection: 'some-collection',
+});
+core.register(gbTrackerPlugin);
+```
+
+It accepts a mandatory `customerId`. It also accepts optional `area` and `collection`
+arguments to handle tracking actions against specific areas and collections.
+
+Additionally, the plugin will automatically listen for and act on beacon
+events. It is designed to work with the other plugins in the Logic Layer.
```
jmbtutor: I don't know if this break was based on a max line length, but I would defer to [semantic line breaks](https://sembr.org/) in this case: join the two lines and break after the period.

packages/@groupby/elements-gb-tracker-plugin/README.md
```diff
@@ -0,0 +1,44 @@
...
+## Events
+
+This plugin listens for a number of events.
+These events are defined in the [`@groupby/elements-events`][elements-events] package.
+
+[elements-events]: https://github.com/groupby/elements-events
+
+### Received
+
+* `BEACON_SEARCH`: When received, a GroupBy `Auto Search` beacon is triggered.
```
jmbtutor: "Auto Search" is not the name of an entity in code, so it shouldn't be in backticks. It can go in double quotes, though, if you still want to set it apart.

```suggestion
* `BEACON_SEARCH`: When received, a GroupBy Auto Search beacon is triggered.
```

packages/@groupby/elements-gb-tracker-plugin/src/gb-tracker-plugin.ts
```diff
@@ -0,0 +1,98 @@
...
+  /**
+   * Returns this plugin's instance of the [[GbTracker]] client.
+   */
+  register(plugins: PluginRegistry): ReturnType<typeof GbTracker> {
+    this.core = plugins;
+    return this.gbTracker;
+  }
+
+  /**
+   * Sets event listeners for beacon events. Also calls the GbTracker method
```
jmbtutor: Calling out `autoSetVisitor` specifically may be a bit too much detail for this doc comment, though that may not be relevant if you take the suggestion below about putting the `autoSetVisitor` call in the constructor.

packages/@groupby/elements-gb-tracker-plugin/src/gb-tracker-plugin.ts
```diff
@@ -0,0 +1,98 @@
...
+      },
+    };
+    this.gbTracker.sendAutoSearchEvent(payload);
+  }
+}
+
+/**
+ * The options used to instantiate a [[GbTrackerPlugin]].
+ */
+export interface TrackerPluginOptions {
```
jmbtutor: If this fits in MVP, you should consider accepting a tracker instance. I'm sure some clients will have already instantiated their own if their site isn't completely on Elements or if they're doing anything complex.

packages/@groupby/elements-gb-tracker-plugin/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "@groupby/elements-gb-tracker-plugin",
+  "version": "0.0.0",
+  "description": "",
+  "main": "dist/index.js",
```
jmbtutor: This is missing the `types` and `esnext` keys. Follow the order found in the other plugins.

DanielMoniz-GBI: I got it from *somewhere*...That implies that the `types` and `esnext` keys are missing elsewhere, or that they were missing and have been fixed.

packages/@groupby/elements-gb-tracker-plugin/src/gb-tracker-plugin.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  /**
+   * A reference to the registry of plugins for later use.
+   */
+  core: PluginRegistry;
+
+  /**
+   * Name of the events plugin.
+   */
+  eventsPluginName = 'dom_events';
```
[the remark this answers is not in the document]
DanielMoniz-GBI: I'm okay with this, although it's worth nothing we'll be pushing a greater inconsistency because three of our plugins still use this convention. I agree that it's not super useful, though.

packages/@groupby/elements-gb-tracker-plugin/src/gb-tracker-plugin.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  /**
+   * The [[GbTrackerPlugin]] constructor instantiates an instance of the
+   * [[GbTracker]] client and attaches it to this plugin's `gbTracker`
+   * property.
+   *
+   * @param options The options for instantiating the GbTracker client.
+   */
+  constructor(options: TrackerPluginOptions) {
```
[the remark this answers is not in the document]
DanielMoniz-GBI: I don't see other plugins destructuring their constructor arguments.

packages/@groupby/elements-gb-tracker-plugin/src/gb-tracker-plugin.ts
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  /**
+   * Sets event listeners for beacon events. Also calls the GbTracker method
+   * `autoSetVisitor()` to allow for future beacon events to be registered.
+   */
+  ready(): void {
+    this.core[this.eventsPluginName].registerListener(BEACON_SEARCH, this.triggerSearchBeacon);
+
+    this.gbTracker.autoSetVisitor();
```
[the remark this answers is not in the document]
DanielMoniz-GBI: I think you're right - this can go in the constructor. I wanted to be sure about timing, but if we've successfully instantiated a tracker, we can obviously call it as well.

packages/@groupby/elements-gb-tracker-plugin/src/gb-tracker-plugin.ts
DanielMoniz-GBI: How will we know how to interact with a client tracker instance? I'm sure we talked about this and I don't remember the nuances. The idea with firing events was to avoid that whole issue could be sidestepped and we can allow clients to use their own trackers by simply listening to events. Or would they pass callbacks for interacting with their tracker?

packages/@groupby/elements-gb-tracker-plugin/src/gb-tracker-plugin.ts
DanielMoniz-GBI: I checked with the Wisdom team. This is apparently for manual testing and was not really intended to be in the package. We will not include it.

packages/@groupby/elements-gb-tracker-plugin/src/gb-tracker-plugin.ts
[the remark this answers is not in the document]
geoffcoutts: The `CachePlugin` shows an example of this.

packages/@groupby/elements-search-driver-plugin/test/unit/common/search-driver-plugin.test.ts
DanielMoniz-GBI: There are, but the overall testing strategy in the file is different and it would be introducing inconsistencies into this file. It's a task to be done, but I think it's tech debt.

packages/@groupby/elements-gb-tracker-plugin/src/gb-tracker-plugin.ts
jmbtutor: We assume that it's an external gb-tracker instance. Instead of creating our own, we allow them to pass one in. It would be a breach of contract if they pass in anything that's not a gb-tracker.

packages/@groupby/elements-gb-tracker-plugin/CHANGELOG.md
```diff
@@ -0,0 +1,10 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased] - [minor]
```
jmbtutor: Formatting.
```suggestion
## [Unreleased] [minor]
```

packages/@groupby/elements-gb-tracker-plugin/package.json
jmbtutor: Looks like it's the SAYT plugin, and apparently we didn't fix it before releasing it.

packages/@groupby/elements-sayt-driver-plugin/CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased] - [minor]
```
jmbtutor:
```suggestion
## [Unreleased] [minor]
```

packages/@groupby/elements-search-driver-plugin/CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased] - [minor]
```
jmbtutor:
```suggestion
## [Unreleased] [minor]
```

packages/@groupby/elements-sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -221,6 +231,15 @@ export default class SaytDriverPlugin<P = Record> implements Plugin {
     return terms.filter((term) => term.value)
       .map((term) => ({ label: term.value }));
   }
+
+  dispatchSearchBeacon(results: Results, originValue: keyof SendableOrigin): void {
```
geoffcoutts: This is missing doc comments.

packages/@groupby/elements-sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -136,7 +140,12 @@ export default class SaytDriverPlugin<P = Record> implements Plugin {
    * @param event Event that contains the Search API request payload.
    */
   fetchProductData(event: CustomEvent<SaytProductsRequestPayload>): void {
```
geoffcoutts: The documentation for this function says that it 

> dispatches events on success and failure. 

Is it worth expanding on that now to define what events those are since they aren't just an error or success event?

geoffcoutts: Turns out the search driver version of this already does this, so it should be updated here too.

packages/@groupby/elements-search-driver-plugin/src/search-driver-plugin.ts
```diff
@@ -92,16 +96,23 @@ export default class SearchDriverPlugin<P = Record> implements Plugin {
    * and emits the result through an event. The result is emitted in a
    * [[SEARCH_RESPONSE]] event. If the search fails for any
    * reason, a [[SEARCH_ERROR]] is dispatched with the error.
+   * Emits a beacon event if search is successful.
```
geoffcoutts:
```suggestion
   * Emits a [[BEACON_SEARCH]] event if search is successful.
```

packages/@groupby/elements-search-driver-plugin/src/search-driver-plugin.ts
```diff
@@ -92,16 +96,23 @@ export default class SearchDriverPlugin<P = Record> implements Plugin {
...
+      query,
+      group,
+      config,
+      origin,
+    } = event.detail;
     this.sendSearchApiRequest({ query, ...config })
-      .then((results) => {
-        const payload: SearchResponsePayload<P> = { ...results, group };
+      .then((response) => {
+        const payload: SearchResponsePayload<P> = { ...response, group };
```
geoffcoutts: For consistency, the SAYT driver still uses `results` in its SAYT and Search calls. Should update those to the same.

packages/@groupby/elements-search-driver-plugin/test/unit/common/search-driver-plugin.test.ts
```diff
@@ -86,20 +91,29 @@ describe('SearchDriverPlugin', () => {
...
-    let results;
+    let response;
     let sendSearchApiRequest;
+    let dispatchSearchBeacon;
+    let core;
+    let query;
 
     beforeEach(() => {
       config = { area, collection };
+      query = 'search term';
```
geoffcoutts: Unless i missed something, this should be fine to put back as a `const` outside of the `beforeEach()`.

packages/@groupby/elements-gb-tracker-plugin/README.md
```diff
@@ -0,0 +1,44 @@
...
+It is designed to work with the other plugins in the Logic Layer.
+
+The plugin registers an instance of the [gb-tracker client](https://www.npmjs.com/package/gb-tracker-client) with Core.
+
+## Events
+
+This plugin listens for a number of events.
+These events are defined in the [`@groupby/elements-events`][elements-events] package.
+
+[elements-events]: https://github.com/groupby/elements-events
```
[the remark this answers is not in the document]
DanielMoniz-GBI: I believe I copied this from another README, so presumably there is another README also doing it wrong. I'll make the change.

packages/@groupby/elements-gb-tracker-plugin/src/gb-tracker-plugin.ts
DanielMoniz-GBI: Ah, I misunderstood. I think we did agree to do that, but I'm going to push it to a follow-up ticket.

packages/@groupby/elements-search-driver-plugin/src/search-driver-plugin.ts
DanielMoniz-GBI: I have updated the matching function in the Sayt Driver, but not the fetchAutocomplete function for reasons of simplicity.

.circleci/config.yml
```diff
@@ -343,6 +354,9 @@ workflows:
       - unit_test_browser:
           name: unit_test_browser_dom_events_plugin
           <<: *requires_build_final
+      - unit_test_browser:
+          name: unit_test_browser_gb_tracker_plugin
+          <<: *requires_build_final
```
DanielMoniz-GBI: Actually, I removed the browser-specific tests, which should remove/undo a couple of changes here.

.circleci/config.yml
```diff
@@ -377,6 +392,7 @@ workflows:
             - unit_test_quickstart
             - unit_test_browser_core
             - unit_test_browser_dom_events_plugin
+            - unit_test_browser_gb_tracker_plugin
```
DanielMoniz-GBI: Same as above - there are no more browser tests for the `GbTrackerPlugin`.

packages/@groupby/elements-gb-tracker-plugin/src/gb-tracker-plugin.ts
```diff
@@ -0,0 +1,98 @@
...
+   * Name of the events plugin.
+   */
+  eventsPluginName = 'dom_events';
+
+  /**
+   * The [[GbTrackerPlugin]] constructor instantiates an instance of the
+   * [[GbTracker]] client and attaches it to this plugin's `gbTracker`
+   * property.
+   *
+   * @param options The options for instantiating the GbTracker client.
```
jmbtutor: Typedoc assigns a destructured parameter a fixed name, which is `_namedParameters`.

```suggestion
   * @param _namedParameters The options for instantiating the [[GbTracker]] client.
```

packages/@groupby/elements-quickstart/src/quick-start.ts
```diff
@@ -45,6 +48,8 @@ export default function quickStart<P>({
     cache_driver,
     /** Options for the DOM Events plugin. */
     dom_events,
+    /** Options for the Gb Tracker plugin. */
```
jmbtutor:
```suggestion
    /** Options for the GB Tracker plugin. */
```

packages/@groupby/elements-quickstart/src/quick-start.ts
```diff
@@ -97,6 +104,8 @@ export interface QuickStartOptions<P> {
     cache_driver?: Partial<CacheDriverOptions>;
     /** Options for the DOM Events plugin. */
     dom_events?: Partial<DomEventsPluginOptions>;
+    /** Options for the Gb Tracker plugin. */
```
jmbtutor:
```suggestion
    /** Options for the GB Tracker plugin. */
```

packages/@groupby/elements-sayt-driver-plugin/CHANGELOG.md
```diff
@@ -4,10 +4,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased] [patch]
+## [Unreleased] [minor]
 ### Fixed
 - ELE-247: Exported the previously hidden `SaytDriverOptions` interface.
 
+### Added
```
jmbtutor: CRAFDS: `Added` goes before `Fixed`.

packages/@groupby/elements-sayt-driver-plugin/README.md
```diff
@@ -24,10 +24,10 @@ core.register(saytDriverPlugin);
...
 This plugin accepts the following options:
 * `productTransformer` (optional)
-    * A function that accepts a Record object and returns a transformed
-    product. This argument is only necessary if the View Layer is being
-    used. More generally, this function is an opportunity to transform
-    the data into a useable format for your application's front-end.
+  * A function that accepts a Record object and returns a transformed
+  product. This argument is only necessary if the View Layer is being
+  used. More generally, this function is an opportunity to transform
+  the data into a useable format for your application's front-end.
```
jmbtutor: These lines should be indented one more level.

packages/@groupby/elements-sayt-driver-plugin/src/sayt-driver-plugin.ts
```diff
@@ -6,6 +8,7 @@ import {
   SAYT_PRODUCTS_REQUEST,
   SAYT_PRODUCTS_RESPONSE,
   SAYT_PRODUCTS_ERROR,
+  BEACON_SEARCH,
```
jmbtutor: Alphabetical order? It doesn't really matter, though.

packages/@groupby/elements-gb-tracker-plugin/test/unit/common/gb-tracker-plugin.test.ts
```diff
@@ -0,0 +1,113 @@
+// eslint-disable-next-line import/no-unresolved
+import * as GbTrackerPackage from 'gb-tracker-client/slim-es';
+import { BEACON_SEARCH } from '@groupby/elements-events';
+import { expect, stub, spy } from '../../utils';
+import GbTrackerPlugin from '../../../src/gb-tracker-plugin';
+
+describe('GbTrackerPlugin', () => {
+  let trackerPlugin: any;
```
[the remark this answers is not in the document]
DanielMoniz-GBI: We generally try to type as generically as possible in our tests. Basically, the idea is that we want the types in our files, but we don't want to spend (sometimes lots of) time worrying about types in our tests.

packages/@groupby/elements-gb-tracker-plugin/src/gb-tracker-plugin.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  /**
+   * Unregister event listeners for beacon events.
+   */
+  unregister(): void {
+    this.core[this.eventsPluginName].unregisterListener(BEACON_SEARCH, this.triggerSearchBeacon);
+  }
+
+  /**
+   * Triggers a Search beacon. Intended for use as an event listener.
```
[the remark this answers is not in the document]
DanielMoniz-GBI: Technically this function acts as a listener when added as one. For example, we might call `addEventListener('click', myFunction)`. Now `myFunction` is registered as an event listener.

PR #46, opened 2019-12-16
ELE-558: Add bundle minification for production bundle

geoffcoutts:
Adding flag for setting production environment in webpack.

Changing the naming scheme of the bundle.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
   ],
   "scripts": {
     "bundle": "webpack",
+    "bundle:prod": "NODE_ENV=production webpack -p",
     "docs": "typedoc --options ./typedoc.js packages",
     "lint:scripts": "eslint ./packages/@groupby/*/src/ --ext .ts --plugin only-warn",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -12,6 +12,7 @@
   ],
   "scripts": {
     "bundle": "webpack",
+    "bundle:prod": "NODE_ENV=production webpack -p",
```
jmbtutor: Is it necessary to specify `NODE_ENV=production`? Shouldn't the `-p` flag already handle that?

webpack.config.js
```diff
@@ -1,18 +1,22 @@
...
     core: path.resolve(__dirname, 'presets/core.ts'),
-    'all-plugins': path.resolve(__dirname, 'presets/all-plugins.ts'),
+    logic: path.resolve(__dirname, 'presets/all-plugins.ts'),
   },
 
   devtool: 'source-map',
 
   output: {
-    filename: 'gbe-[name]-bundle.js',
+    filename: () => {
```
jmbtutor: This can use the expression form (i.e. no braces).

